PR #106, opened 2025-04-10
Add missing Plausible events

Pabl0cks:
Adding some Plausible events that were missing:

- "joinBG" event after a successful response Join BuidlGuidl action
- "signupSRE" event after a successful response of registration to SRE action
- "challengeSubmission" after a successful response from a challenge submission

Each event is tracked in the `onSuccess` callback, let me know if this isn't the best approach / if server-side tracking would be preferable for any of these events


carletex (changes requested):
Hey @Pabl0cks thanks for this! This should work, but since we have the option here, it makes more sense to have them on the backend.

Having it on the front-end has 2 issues:
1. Blocked by Brave / Ad Blocker / etc
2. People leaving the page (or refresh) before the onSuccess is called (but the request has been made)

Pabl0cks:
Thanks for the review @carletex, it def makes more sense in the backend, just updated it! 🙏

Should we implement some check to not store stats from testing environments?

packages/nextjs/app/api/challenges/[challengeId]/submit/route.ts
```diff
@@ -64,6 +65,8 @@ export async function POST(req: NextRequest, { params }: { params: { challengeId
       return NextResponse.json({ error: "Failed to create submission" }, { status: 500 });
     }
 
+    await trackPlausibleEvent("challengeSubmission", { challengeId }, req);
```
Comment (technophile-04) [inline packages/nextjs/app/api/challenges/[challengeId]/submit/route.ts:68]: Ohh maybe we wrap this inside waitUntil too? And at other files too? 

Or lol maybe I am over optimising stuff becoz on preview link I tried signing up and it was quick!

Comment (carletex) [reply packages/nextjs/app/api/challenges/[challengeId]/submit/route.ts:68]: On BG3.5 I was removing the `await`, since we didn't care about the result or it failed.... so it minimally impacted the response time.

This might not be an option on Vercel serverless functions, right, Shiv? Since Vercel might halt/kill execution after returning.

If that's the case, I think moving them into `waitUntil` is a great idea :)

packages/nextjs/services/plausible/index.ts
carletex, line 1: I think we don't need a folder for this since it'd be a simple service (as github.ts, autograder.ts, etc). It could just be `plausible.ts` in the upper folder

packages/nextjs/services/plausible/index.ts
```diff
@@ -0,0 +1,33 @@
+const PLAUSIBLE_EVENT_ENDPOINT = "https://plausible.io/api/event";
+
+export async function trackPlausibleEvent(
+  eventName: string,
```
carletex, line 4: I think we could do a ENUM / TYPE here instead of string. So it helps us to keep track of the events and be more type safe.

e.g.

```
export enum PlausibleEvent {
  JOIN_BG = "joinBG",
  SIGNUP_SRE = "signupSRE",
  CHALLENGE_SUBMISSION = "challengeSubmission",
}
```

and we use them as `PlausibleEvent.JOIN_BG`

packages/nextjs/app/api/challenges/[challengeId]/submit/route.ts
Comment (technophile-04) [reply packages/nextjs/app/api/challenges/[challengeId]/submit/route.ts:68]: > On BG3.5 I was removing the await, since we didn't care about the result or it failed.... so it minimally impacted the response time.

> This might not be an option on Vercel serverless functions, right, Shiv? Since Vercel might halt/kill execution after returning.

I remember nicely that vercel terminates the fucntion after response is sent lol not able to find the perfect reference of it but for sure I remember reading it but I might be halucinating too 😅.  Since doing the research I am not able to find the reference now, but for sure it timesout in 15sec. So we might or might not get the analytics if we don't have `await`. 

But here https://vercel.com/changelog/waituntil-is-now-available-for-vercel-functions 

> It's used to run anything that can be done after the response is sent, such as logging, sending analytics, or updating a cache, without blocking the response from being sent. 

They suggest using `waitUnitl ` for sending analytics

Pabl0cks:
Thanks @carletex and @technophile-04 for all the suggestions and insights ♥ 

I hope I have addressed all the requested changes in https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/106/commits/eb699036bf962b7d6de8343143c95ef1ecb6c459 🙌

packages/nextjs/services/plausible.ts
```diff
@@ -0,0 +1,41 @@
...
+  waitUntil(
+    fetch(PLAUSIBLE_EVENT_ENDPOINT, {
+      method: "POST",
+      headers,
+      body: JSON.stringify(payload),
+    }).catch(error => {
+      console.error("Error tracking Plausible event:", error);
+    }),
+  );
+}
```
carletex, line 41: `waitUntil` is tied to the backend function, so it's better to add it there (some backend functions already have the waitUntil).... and decouple this service/util from it.

packages/nextjs/services/plausible.ts
```diff
@@ -0,0 +1,41 @@
+import { waitUntil } from "@vercel/functions";
+
+export enum PlausibleEvent {
+  JOIN_BG = "joinBG",
+  SIGNUP_SRE = "signupSRE",
+  CHALLENGE_SUBMISSION = "challengeSubmission",
+}
```
carletex, line 7: @technophile-04 suggested this to Pablo, but I've never used it. Not sure what is the PROS/CONS (vs. just a `const` or `as const` + type from it)

technophile-04: Screenshot 2025-04-13 at 1 10 50 PM

The best part about `enum` is that developer can't pass the literal string even if its valid. Like TS gives the error and you have to always use `PlausibleEvent.{value}`. 

The con is only that `enum` are not actual things in JS and just TS utility so when it get compiled it does some hacky tricks that's why people don't like `enum` since it adds a bit runtime overhead. 

<details>

<summary>
    example compiled enum
</summary>

```js
    var PlausibleEvent;
    (function (PlausibleEvent) {
        PlausibleEvent["JOIN_BG"] = "joinBG";
        PlausibleEvent["SIGNUP_SRE"] = "signupSRE";
        PlausibleEvent["CHALLENGE_SUBMISSION"] = "challengeSubmission";
    })(PlausibleEvent || (PlausibleEvent = {}));
```

</details>

But I think in our case it's fine? Since it's a not big deal and also we have been using the enum in 

https://github.com/BuidlGuidl/SpeedRunEthereum-v2/blob/main/packages/nextjs/services/database/config/types.ts

packages/nextjs/services/plausible.ts
Pabl0cks: Updated in https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/106/commits/644eca28fd544a0826ec3a8768422d578435ba6d 🙏

packages/nextjs/services/plausible.ts
carletex: thanks Shiv! Got it.

carletex (approved):

PR #107, opened 2025-04-10
Mobile Header and Table Tweeks

ChangoMan:
Adjusting the column width on the tables. For the mobile header, I don't think the burger menu is needed (at least right now).

The only link in the burger menu showing up is `Portfolio` which is also under the `AddressInfoDropdown` menu. If we add the logo on mobile that links to the homepage, there isn't really a need for the burger menu.

If more links are needed, we could just put them under the `AddressInfoDropdown` menu. Let me know if this works. I did mess with just showing the "running girl" and having both burger and connect button, but it did not look great.

Screenshot 2025-04-10 at 12 35 54 PM

Screenshot 2025-04-10 at 12 36 06 PM


carletex (approved):
Great stuff @ChangoMan 

> If more links are needed, we could just put them under the AddressInfoDropdown menu.

100%. I completly missed this: we already have a _burger_ menu!

PR #110, opened 2025-04-10
Adding Not Found Component

ChangoMan:
Made a component out of the not found page, re-use it for a user not found.

Screenshot 2025-04-10 at 1 55 12 PM

### DIFF
--- a/packages/nextjs/app/builders/[address]/page.tsx
+++ b/packages/nextjs/app/builders/[address]/page.tsx
@@ -1,7 +1,7 @@
+import { notFound } from "next/navigation";
 import { UpgradedToBGCard } from "./_components/UpgradedToBGCard";
 import { UserChallengesTable } from "./_components/UserChallengesTable";
 import { UserProfileCard } from "./_components/UserProfileCard";
-import { NotFound } from "~~/components/NotFound";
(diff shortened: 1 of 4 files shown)


packages/nextjs/app/builders/[address]/page.tsx
```diff
@@ -13,7 +14,17 @@ export default async function BuilderPage({ params }: { params: { address: strin
...
+      <NotFound
+        heading="User Not Found"
+        extraElement={
+          <>
+            <p className="mb-0">Double check the Ethereum address is correct in the URL</p>
+            <p className="m-0">Make sure not to use an ENS address in the URL</p>
+          </>
+        }
+      />
+    );
```
Comment (carletex) [inline packages/nextjs/app/builders/[address]/page.tsx:27]: The problem with this approach is that we are getting a 200 (this is bad for indexing)

image

Is it possible to return a 404 with a customized page? (as `notFound()`)

Comment (ChangoMan) [reply packages/nextjs/app/builders/[address]/page.tsx]: Ah yes, I added a new `not-found.tsx` inside of the `/builders/[address]` folder. This will make it custom for this route

Comment (carletex) [reply packages/nextjs/app/builders/[address]/page.tsx]: Great stuff!

carletex (approved):
Thanks Hunter!!

PR #111, opened 2025-04-10
Only allow seed on localhost

carletex:
Deleting data on seed is scary; so better to only allow it on localhost... just in case :)


Pabl0cks (approved):
Working great! Good catch preventing this possible future pain 🙌

PR #112, opened 2025-04-11
Improve seed workflow

carletex:
Not sure where, but we used this pattern before, and it was useful.

The goal is for all of us to have a customized seed: if not, I always have the seed data modified (git tracked file) to accommodate the data I want.

This way, we add an extra step (copy example to seed.data), but you can add and keep the data that you want. I also added a version check in case we do schema changes (so seed example should change). 

So if you do a PR that changes the schema:
- Update schema file
- Update seed example file + version with new data (updated to the right schema, e.g. we add a field)

Let me know if this makes sense!


packages/nextjs/services/database/seed.ts
```diff
@@ -1,19 +1,65 @@
...
+    // @ts-ignore: seed.data.ts is gitignored and may not exist
+    const seedData = (await import("./seed.data")) as SeedData;
+    const exampleData = (await import("./seed.data.example")) as SeedData;
+
+    if (seedData.SEED_DATA_VERSION !== exampleData.SEED_DATA_VERSION) {
+      console.error(
+        `Error: Version mismatch between seed.data.ts (${seedData.SEED_DATA_VERSION}) and seed.data.example.ts (${exampleData.SEED_DATA_VERSION})`,
+      );
+      process.exit(1);
+    }
```
technophile-04, line 36: Love this idea of version mistmatch!! 

image

packages/nextjs/.gitignore
```diff
@@ -34,4 +34,7 @@ yarn-error.log*
 .env.production.local
 
 # typescript
-*.tsbuildinfo
\ No newline at end of file
+*.tsbuildinfo
+
+# database
+services/database/seed.data.ts
```
technophile-04, line 40: Ohh we need to remove current seed.data.ts file : 

```
git rm -r --cached packages/nextjs/services/database/seed.data.ts
``` 

And then commit it 🙌

technophile-04 (approved):
Works nicely!! Just added a small comment, Thanks Carlos makes a lot of sense and its great!

packages/nextjs/.gitignore
carletex: Oops, missed that. Thanks!! https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/112/commits/c82d0ef64c4ce61873982de9f8dce9b4a77ee7a2

carletex: Oh, It's detecting it as moving. It was confusing (no deleted files) but it should be good!

image

rin-st (approved):
Like!

PR #46, opened 2025-03-13
[WIP] Data migration process

Pabl0cks:
**This PR is not intended to be merged into [main] but serves as a testing ground for the data migration process.**

----------------------------------------------------------------

I've tried to follow the same data structure and script that we used for our seed sample import. **This PR contains:**

- SRE data export generated on March 12, using `https://scaffold-directory-dev.ew.r.appspot.com/builders` as the source of truth to generate the data. Data is splitted by data type to not have a 80mb unique file.
  - `productionUsers.data.ts`
  - `productionChallenges.data.ts`
  - `productionUserChallenges.data.ts`
  - `productionEvents.data.ts`
- New script `importProduction.ts` for importing production data

### How to test
Run the import script from the nextjs folder:
```
cd packages/nextjs && yarn tsx services/database/importProduction.ts
```

You can compare users from https://speedrunethereum.com/builders to the imported data and verify if everything matches. I've used DBeaver for the data verification, but probably we can merge soon the profile page into this branch and we could verify directly on http://localhost:3000/

### Required steps before actual migration

- Merge [main] changes to this branch and update the script to adapt any database schema changes (e.g., socialX field name changes that are pending to merge)
- Re-generate data scripts with fresh data using the export script from the [sre-export](https://github.com/carletex/bg-scripts/tree/sre-export/scripts/sre) branch in our `bg-scripts` repository
- Configure the SRE-v2 production database as target and execute

### Notes
- The script uses batched inserts to handle large datasets, was getting some errors otherwise
- Statistics are provided at the end to verify the import was successful

Closes #19 
Closes #4

PR #117, opened 2025-04-11
drizzle migration flow

technophile-04:
### Some research: 

There are two types of migration approach: 

1. Database first: when your database schema is a source of truth. So we pull the schema by connecting to database and generating the `schema.ts` file in typescript. 

2. Codebase first: Where the database schema lives in codebase and push it to database.  

In Codebase first approach drizzle provides two kind of workflows: 

1. `drizzle-kit push`: What it does is instead of generating the intermediary `.sql` file it just pushes the stuff directly to database having source of truth as `scheam.ts` file. 

2. `drizzle-kit generate && dirzzle-kit migrate`: 

The `drizzle-kit generate`: will generate the intermediate `.sql` migration file based of current `schema.ts` 

The `drizzle-kit migrate`: will the pull the database make sure everything is ok and push it to the database using the `.sql` file generate. 

<details>

<summary>
    nice visual from drizzle docs
</summary>

image

</details>

So we have been using codebase first approach only, but earlier we were using push method and now moving to 2nd method . 

---

### Description: 

So earlier we were using codebase first approach 1st part it with push and now we are switching to migration workflow. 

So the way the migration workflow works internally is along with the things mentioned in above research section, when we run `drizzle-kit migrate` the drizzle also maintain a table in db called `__drizzle_migrations` table to track migration along with git. 

<details>

<summary>
    drizzle migrations table from local studio
</summary>

Screenshot 2025-04-11 at 2 50 27 PM

</details>

Lol ideally before pushing to prod we should have switched to migration workflow from the beginning only, because now when I tried switching to migration workflow I was getting this : 

<details>

<summary>
    error image enum already exist
</summary>

Screenshot 2025-04-11 at 12 58 46 PM

</details>

But I worked around this solution by using [below (one time) script](https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/117/commits/9e5599359aef7dded74d02acd714d1e0e2f2ae31), which helps us switching from push to migration flow, found it in this discussion https://github.com/drizzle-team/drizzle-orm/discussions/1604

And now I ran `yarn drizzle-kit migrate` everything works nicely, so pushed it to prod db to and now we have their `__drizzle_migration` table too  🙌  (already ran `yarn drizzle-kit migrate` command in prod db) 

Also in local db, I added a new field to generate the migration and pushed it it was also working nicely 🙌


README.md
```diff
@@ -55,3 +55,19 @@ To iterate on the database:
...
+yarn drizzle-kit generate
+```
+
+Then we can apply the migration with:
+
+```
+yarn drizzle-kit migrate
+```
+
+We also need to make sure we commit the migration to the repo.
```
technophile-04, line 73: Happy to update this in other way, if we want 🙌

carletex: Should we tweak the order here (first migrations, second "push")

If someone is setting the environment, I guess we want them to use `migrate` instead of `push`, right?

(and later on, we can hint that you can use `push` to tinker, and then run `generate` to create the migration).

Let me know if you need help.

Thanks!

carletex:
This is amazing, Shiv! Thanks for the explanation. The script you ran seems non-destructive so it was ok to YOLO it :)

Just left a tiny comment for the documentation.

README.md
technophile-04: Ohhh yess!!! Just pushed some tweaks but please let me know if I should tweak more 🙌

technophile-04:
> The script you ran seems non-destructive so it was ok to YOLO it :)

Yeah lool I ran that script 3-4 time locally with #46 data to make sure it doesn't destroy any data, but totally there was still 1% chance of something going wrong with prod and me sitting out of BG 😂

carletex (approved):
Amazing, this is ready!

one question:

image

There is no way around this incomplete feedback, right?

If I destroy the local database and do a `migrate`, I get the same feedback as if I run the migration in an already up-to-date database.

I remember seeing which migrations were applying, but that maybe was Prima (and I also remember talking about this in other project :))

technophile-04:
> If I destroy the local database and do a migrate, I get the same feedback as if I run the migration in an already up-to-date database.

> I remember seeing which migrations were applying, but that maybe was Prima (and I also remember 
talking about this in other project :))

yeah drizzle only give this minimal logs :( and there is no way we can make them more verbose, also there is an issue related to this https://github.com/drizzle-team/drizzle-orm/issues/3410 

And also they have huge discussion on migration update process [here](https://github.com/drizzle-team/drizzle-orm/discussions/2624)

rin-st (approved):
Works great! Good job, thank you!

PR #119, opened 2025-04-11
Add GH action to store a daily database backup in AWS

Pabl0cks:
First step to tackle #5 

This PR adds a GH action to make a daily backup from neon database and store it in a AWS S3 bucket. 

This will be complementary to the [native Instant Restore](https://neon.tech/docs/manage/backups) from Neon (1 to 7 days retention depending if we upgrade to the first paid plan), which seems to be better for "hot" restores, since it seems you can instantly restore a db snapshot if you need to execute a rollback.

I followed the guide from the previous link and tested it with staging database in my [local repo](https://github.com/Pabl0cks/sre-aws-backups), verifying that I could restore the database and could access all the data locally.

In this repo I changed the `PG_VERSION` from 17 (staging) to 15 (prod), and we'd need to set the following environment variables in this repo:

- `DATABASE_URL`
- `IAM_ROLE`
- `AWS_ACCOUNT_ID`
- `S3_BUCKET_NAME`

`DATABASE_URL` maybe is already set or with a different name, I can tweak it in the script if so. Can send you the rest of the values in a private msg.

Once set, we should be able to execute it from the "Actions" tab and the backup should appear in our AWS S3 backup, under the folder with the action name ("daily-db-backup" in this case).

### DIFF
--- a/.github/workflows/daily-db-backup.yml
+++ b/.github/workflows/daily-db-backup.yml
@@ -1,0 +1,69 @@
+name: daily-db-backup
+
+on:
+  schedule:
+    - cron: "0 5 * * *" # Runs at 5 am ET (us-east-1)
+  workflow_dispatch:
+
+jobs:
+  db-backup:
+    runs-on: ubuntu-latest
+
+    permissions:
+      id-token: write
+
+    env:
+      RETENTION: 7
+      DATABASE_URL: ${{ secrets.DATABASE_URL }}
+      IAM_ROLE: ${{ secrets.IAM_ROLE }}
+      AWS_ACCOUNT_ID: ${{ secrets.AWS_ACCOUNT_ID }}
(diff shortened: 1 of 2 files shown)


carletex:
Hey Pablo, good stuff, thanks!!

Maybe we can pair next week and we set up all the env vars and test it.

Some comments/quetions:
- Love the hot rollaback on Neon. Will probably ask Zak to upgrade the account.
- With a non-free Neon account, we can also have read-only keys (which it'll be great for this use case)
- Maybe daily backup is overkill if we have 7 day data retention?

> verifying that I could restore the database and could access all the data locally.

What did you do to restore the database? Should we add some of this info in the README (or create a new `contributing` / `developing` md file)

carletex (approved):

carletex (approved):

PR #120, opened 2025-04-11
Adding Custom Builder OG Image

ChangoMan:
Generate a custom OG image for the `builders/[address]` page. Most of the code taken from https://github.com/technophile-04/sendeth

Question: do we want to keep the `Scan to send ETH` part? 

Visit the api endpoint to test out the image:
```
http://localhost:3000/api/og?address=0x34aA3F359A9D614239015126635CE7732c18fDF3
```

Screenshot 2025-04-11 at 3 18 18 PM


carletex:
Thanks for Tinkering with this, Hunter!

> Question: do we want to keep the Scan to send ETH part?

Probably not, right?

One initial iteration could be:
- SRE logo
- User punkblockie (future: ens avatar if exist)
- User ENS (if exist) + short form address
- Challenge completed: X
- Join date? (not sure about this one)

If we make it obvious, Maybe we can remove the "Builder Profile" copy too.

ChangoMan:
> One initial iteration could be:
> 
> * SRE logo
> * User punkblockie (future: ens avatar if exist)
> * User ENS (if exist) + short form address
> * Challenge completed: X
> * Join date? (not sure about this one)
> 
> If we make it obvious, Maybe we can remove the "Builder Profile" copy too.

Ok I've adjusted the image, removed the QR code, and added Challenges completed. I added the ENS avatar if available, otherwise it will use the "Punk Blockie". The Punk image does look a little blurry though, I assume it's not suppose to be very big.

With ENS Avatar:
Screenshot 2025-04-14 at 3 58 07 PM

With Punkblockie:
Screenshot 2025-04-14 at 3 58 03 PM

packages/nextjs/app/api/og/route.tsx
```diff
@@ -0,0 +1,269 @@
...
+    } catch (error) {
+      console.error("Error resolving ENS name:", error);
+    }
+
+    let challenges: UserChallenges = [];
+    try {
+      challenges = await getLatestSubmissionPerChallengeByUser(address);
+    } catch (error) {
+      console.error("Error fetching builder challenges", error);
+    }
```
carletex, line 45: We need to filter out the challenges that are not ACCEPTED. Check `enum ReviewAction`

packages/nextjs/app/api/og/route.tsx
```diff
@@ -0,0 +1,269 @@
...
+                marginTop: 5,
+                marginBottom: 16,
+                display: "flex",
+                fontSize: 36,
+                fontWeight: "bold",
+                color: colors.text,
+                textAlign: "center",
+              }}
+            >
+              Challenges Completed: {challenges.length}
```
carletex, line 253: I think these two _compete_ too much
image

Maybe we need one of the two to be bigger / smaller or bold / light?

packages/nextjs/app/builders/[address]/page.tsx
```diff
@@ -2,11 +2,59 @@ import { notFound } from "next/navigation";
...
+    address: string;
+  };
+};
+
+export async function generateMetadata({ params }: Props): Promise<Metadata> {
+  const address = params.address;
+  const isValidAddress = isAddress(address);
+
+  // Default title and description
+  const title = "Builder Profile | Speed Run Ethereum";
```
Comment (carletex) [inline packages/nextjs/app/builders/[address]/page.tsx:23]: Could the title be dynamic too? `ens or short address | Speed Run Ethereum`

maybe we could return this data from the og endpoint too (in case you don't want to compute twice)

e.g. `const [addressOrEns, ogImageUrl] = /api..`

carletex (commented):
This is looking great! Added a few more comments.

I think it could also be more ornamental (with the girl running on the bottom or similar) but not sure how to describe it. So happy to merge this as a first iteration.

packages/nextjs/app/api/og/route.tsx
ChangoMan: Ah thanks, fixed via https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/120/commits/645ebd5768a70932d08dc034588557781c2c8e1f

packages/nextjs/app/api/og/route.tsx
ChangoMan: I've adjusted the font sizes and colors, it doesn't seem like font weight is working though

packages/nextjs/app/builders/[address]/page.tsx
Comment (ChangoMan) [reply packages/nextjs/app/builders/[address]/page.tsx]: I was able to make the title dynamic, I split out a new function for getting the ENS or Short Address at `packages/nextjs/utils/ens-or-address.ts`

I did try to have the og endpoint return both, but it resulted in the image not working.

packages/nextjs/app/api/og/route.tsx
carletex: This looks great!! <3

Will do a final pass tomorrow.

Thanks Hunter!

packages/nextjs/utils/ens-or-address.ts
```diff
@@ -0,0 +1,26 @@
+import { Address, createPublicClient, http } from "viem";
+import { mainnet } from "viem/chains";
+import { getAlchemyHttpUrl } from "~~/utils/scaffold-eth";
+
+const alchemyHttpUrl = getAlchemyHttpUrl(mainnet.id);
+const rpcUrl = alchemyHttpUrl;
```
technophile-04, line 6: nit: we can directly name alchemyHttpUrl => rpcUrl? or just pass it http

packages/nextjs/app/builders/[address]/page.tsx
Comment (technophile-04) [reply packages/nextjs/app/builders/[address]/page.tsx]: Ohh yeah :( dind't find the straigt forward way to make both working

technophile-04 (approved):
Thanks @ChangoMan!! Its looking really nice and great first iteration!! 

maybe we can explore in new PR's how to make punkblocie less blurrry but looks soo nice with ENS image, Thanks!! 

also it works nicely, I think in preview on vercel open graph it's not working because their `VERCEL_PROJECT_PRODUCTION_URL` is `speedrunethereum.com` and not able to fetch from endpoint : 

Screenshot 2025-04-17 at 5 43 56 PM

But I think once we merge this will be working nicely 🙌

carletex (approved):

carletex:
It seems that something doesn't work on Vercel Production:

https://speedrunethereum.com/api/og?address=0x1c131c894ec4d4E1040d9811328158DBacde5AC3

Issue #125, opened 2025-04-14
Batches: CRUD


carletex:
We'd need #122 first. 

## 1. Admin page : CRUD for batches

(let's not have the "delete" for now)

![Image](https://github.com/user-attachments/assets/c3abd7be-7150-46ef-a608-f966ad321059)

![Image](https://github.com/user-attachments/assets/dfe2c05a-ced9-4c84-8bcf-0393983ef927)

## 2. Builders in Batches

It shows all the users that belong to a batch.

![Image](https://github.com/user-attachments/assets/c76ff27e-6f50-469f-89b0-0703fba2816c)

We should be able to switch their "status"

![Image](https://github.com/user-attachments/assets/3690a190-29e2-4de8-b344-6c3955978429)

Note: In BG.3.5, they get tagged as candidates when they sign a message to request access to the Telegram channel (check #124). Then, we manually mark them as graduated (in the future we could automate this)

carletex:
Hey @rin-st , let me know if you need more info! Also feel free to split in 2 PRs if you think that's more appropriate.

I'll be creating an initial PR for #122 so you can branch from it.

rin-st:
Hey @phipsae, I think it will be easier if I take it since I already worked on the first part. Thanks!

phipsae:
Make sense, just wanted to support somehow.

rin-st:
#137 and #154

Issue #113, opened 2025-04-11
[TRACKING] Breaking the SRE => BG3.5 link


carletex:
The current flow is like:

Builders tackle SRE first few challenges => join the BG => join a batch / update profile / submit builds. etc.

Moving forward, we want to make SRE the central piece (like the "Ethereum Builder Profile"), so we don't want people to join the BG app. To do so, we need to tackle a bunch of things. (This issue is just going to be a tracking issue; we'll be creating several PRs/issues).

Created the first round of sub-issues.

---

Am I missing something? Let's add it here!

carletex:
Just created a few sub-issues for this.

### PR workflows

There are a bunch of things to tackle (in this big issue, but this applies to every PR): let's avoid big PRs.

- Let's use `feature` branch in cases where we can't merge to main (because it'd break since something might be missing)
- Let's merge to main even if the feature is not completed (e.g. we could merge some batch schema tweaks to main, even if we are not using the data anywhere).

PR #126, opened 2025-04-14
Batches: schema

carletex:
Added the needed schema + seed data to allow Batches.

Went with 1:1 relationship between Users and Batches, since it's what we are doing right now, and it was the simplest approach. Since these are onboarding batches, people should only do one. Sometimes a person can do the batch because they are busy, and we move them to the next batch (and we don't need to keep a registry of it).

Let me know if I'm missing something and feel free to branch from this PR if you are working on batches stuff.

---

Also, PR workflow from: https://github.com/BuidlGuidl/SpeedRunEthereum-v2/issues/113#issuecomment-2796081557

> There are a bunch of things to tackle (in this big issue, but this applies to every PR): let's avoid big PRs.
> - Let's use feature branch in cases where we can't merge to main (because it'd break since something might be missing)
> - Let's merge to main even if the feature is not completed (e.g. we could merge some batch schema tweaks to main, even if we are not using the data anywhere).

Not saying that we need to merge this right now, but after we test it a bit (e.g. with #125) maybe we can merge to avoid a big fat PR at the end.

This is also a good opportunity to see how we work with a staging database or production database (with Neon branches)

----

To test, usual workflow with the database `yarn drizzle-kit migrate` + `yarn db:seed` (you'll need to update your seed data file ;))

### DIFF
--- a/packages/nextjs/services/database/config/schema.ts
+++ b/packages/nextjs/services/database/config/schema.ts
@@ -76,6 +76,7 @@ export const batches = pgTable("batches", {
   name: varchar({ length: 255 }).notNull(),
   startDate: timestamp().notNull(),
   status: batchStatusEnum().notNull(),
+  contractAddress: varchar({ length: 42 }),
   telegramLink: varchar({ length: 255 }),
 });
 
--- a/packages/nextjs/services/database/config/schema.ts
+++ b/packages/nextjs/services/database/config/schema.ts
@@ -1,4 +1,4 @@
-import { ReviewAction, UserRole } from "./types";
+import { BatchStatus, BatchUserStatus, ReviewAction, UserRole } from "./types";
 import { SQL, relations, sql } from "drizzle-orm";
 import {
   AnyPgColumn,
@@ -27,11 +27,17 @@ export const reviewActionEnum = pgEnum("review_action_enum", [
 
 export const userRoleEnum = pgEnum("user_role_enum", [UserRole.USER, UserRole.BUILDER, UserRole.ADMIN]);
 
+export const batchStatusEnum = pgEnum("batch_status_enum", [BatchStatus.CLOSED, BatchStatus.OPEN]);
+export const batchUserStatusEnum = pgEnum("batch_user_status_enum", [
+  BatchUserStatus.GRADUATE,
+  BatchUserStatus.CANDIDATE,
+]);
+
 export const users = pgTable(
   "users",
   {
     userAddress: varchar({ length: 42 }).primaryKey(), // Ethereum wallet address
-    role: userRoleEnum().default(UserRole.USER), // Using the enum and setting default
+    role: userRoleEnum().default(UserRole.USER),
     createdAt: timestamp().defaultNow().notNull(),
     updatedAt: timestamp().defaultNow().notNull(),
     socialTelegram: varchar({ length: 255 }),
@@ -40,6 +46,8 @@ export const users = pgTable(
     socialInstagram: varchar({ length: 255 }),
     socialDiscord: varchar({ length: 255 }),
     socialEmail: varchar({ length: 255 }),
+    batchId: integer().references(() => batches.id),
+    batchStatus: batchUserStatusEnum(),
   },
   table => [uniqueIndex("idUniqueIndex").on(lower(table.userAddress))],
 );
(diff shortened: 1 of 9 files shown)


packages/nextjs/services/database/config/schema.ts
```diff
@@ -50,19 +58,28 @@ type ExternalLink = Partial<{
...
   disabled: boolean().default(false),
   previewImage: varchar({ length: 255 }),
   icon: varchar({ length: 255 }),
   dependencies: varchar({ length: 255 }).notNull().array(),
   externalLink: jsonb("external_link").$type<ExternalLink>(),
 });
 
+export const batches = pgTable("batches", {
+  id: serial().primaryKey(),
+  name: varchar({ length: 255 }).notNull(),
```
carletex, line 76: In BG3.5 we are saving it with random UUID as id and name as "1", "2", etc.

Here I'm using "serial" (since we are doing it in the userChallenges table) + name will be "Batch 1", etc (instead of just "1") so it allows some flexibility (if we do a batch for univeristy people, it might be call different). It'd be like a display name (and we can always sort by startDate or ID if serial)

What do you think?

packages/nextjs/services/database/config/schema.ts
```diff
@@ -40,6 +46,8 @@ export const users = pgTable(
     socialInstagram: varchar({ length: 255 }),
     socialDiscord: varchar({ length: 255 }),
     socialEmail: varchar({ length: 255 }),
+    batchId: integer().references(() => batches.id),
+    batchStatus: batchUserStatusEnum(),
```
carletex, line 50: As I mentioned in the PR description, 1:1 relation

rin-st:
> This is also a good opportunity to see how we work with a staging database or production database (with Neon branches)

Where can I read more about this?

packages/nextjs/services/database/config/schema.ts
technophile-04: Sounds good to me!

carletex:
> Where can I read more about this?

https://neon.tech/docs/introduction/branching

The ideal scenario (let's not do it for now since it might be overkill):
1. You create a PR
2. Neon automatically creates a database branch
  - It runs `migrate`
  - seed: You can select "production" data o "seed" data to feed the database
4. Your preview Vercel points to the created database branch

For now, maybe we could do it manually: manually create a branch for a PR (e.g. batches) and import the data / tweak schema as needed... then update POSTGRES_URL for my preview branch.

packages/nextjs/services/database/config/schema.ts
technophile-04: I think we also need to add the relations: 

```ts
export const usersRelations = relations(users, ({ many, one }) => ({
  userChallenges: many(userChallenges),
  batch: one(batches, {
    fields: [users.batchId],
    references: [batches.id],
  }),
}));

export const batchesRelations = relations(batches, ({ one }) => ({
  users: one(users, {
    fields: [batches.id],
    references: [users.batchId],
  }),
}));
```
This will make sure when we query get proper types, if we don't add then it errors out while querying 🙌

image

carletex: Oh, I omitted that because I thought those were for 1:N / N:N relations, but now I see that it's for every relation! And it's only for types, etc (migration files are still the same!)

Thank Shiv! 336a8e6d1f58

technophile-04:
Thanks Carlos!! looking good to me! Just added a small comment 

> For now, maybe we could do it manually: manually create a branch for a PR (e.g. batches) and import the data / tweak schema as needed... then update POSTGRES_URL for my preview branch.

Ohh makes sense! Lol also love the above approach too but true it overkill creating database of each branch/PR. 

Also love the idea of having `feature` / `staging` branch? Maybe we keep merging some DB changes to that branch `staging`. And create a final PR ot staging to `main` ? 

And in neon we create a staging db, and configure vercel such that any PR pointed to `staging` branch use that staging databse?  

Just thinking out loud will try to dig in more 🙌

carletex:
> Just thinking out loud will try to dig in more 🙌

Yes!

I think the manual flow works great:
1. Create the branch database on Neon
2. Run migrations / import data to that branch (from local)
3. Point to that database branch (POSTGRES_URL env var) on Vercel for YOUR specific branch (just saw that you can do it, check ss)

image

 let's keep the conversation going (maybe in another issue). Let's find a good balance.

> Thanks Carlos!! looking good to me! Just added a small comment

Done!

I'd be happy if we merge this even if we don't still use it. But happy to wait in case we want to test usecase a bit more.

carletex:
This PR should now be connected to the database branch `batches` (ran the migration too)

image

technophile-04 (approved):
Thanks!! looks great to me! 

> I'd be happy if we merge this even if we don't still use it. But happy to wait in case we want to test use cases a bit more.

ohhh yesss, I think we merge this 🙌 will also resolve migration conflict then in #127

Issue #97, opened 2025-04-07
Import socials and location from bg3.5


Pabl0cks:
To tackle after we launch v2:

- Create location field in v2 and show it in SREv2 profile
- Import socials and location from bg3.5 database
- Hide location from 3.5? Or update data source and hide update button

Issue #123, opened 2025-04-14
Batches: Import script from BG3.5


carletex:
Import batch data from BG3.5:

- Batch info
- builder-batch info

carletex:
Fixed on #136

PR #127, opened 2025-04-15
allow adding location in builders page

technophile-04:
### Description: 

First part of #97, this allows builders adding locations and we store them in SREv2 DB. 

Created the migration, before merging to `main` we need to run `yarn drizzle-kit migrate`

https://github.com/user-attachments/assets/38c66c81-a2a5-4e17-b6ef-1c297e14aacc

PS: When we merge this in `feature` or any branch we are gonna have migration conflicts because of #126, 

here is the discussion regarding it: https://github.com/drizzle-team/drizzle-orm/discussions/1104#discussioncomment-10814609 

The developer could resolve the conflict locally but they also suggested generate the migration file only in CI so while local development we use `drizzle-kit push` and in CI when we merge to `main` it generate the migration files

### TOOD: 

Create a script which udpates the current builders location


packages/nextjs/app/builders/[address]/_components/UpdateLocationModal.tsx
```diff
@@ -0,0 +1,84 @@
...
+      const handleKeyDown = (e: KeyboardEvent) => {
+        if (e.key === "Escape") {
+          closeModal();
+        }
+      };
+      window.addEventListener("keydown", handleKeyDown);
+      return () => {
+        window.removeEventListener("keydown", handleKeyDown);
+      };
+    }, [isOpen, closeModal]);
```
Comment (technophile-04) [inline packages/nextjs/app/builders/[address]/_components/UpdateLocationModal.tsx:30]: Not using `daisyUI` modal here since the using it, it uses `dialo` element which causes the `react-flags-select` render very weirdly and hard to style since dialog is put on `top-layer` and having z-index or anything to ReactFlags selsect was not working so had to implement own modal. 

Tried keeping it very simillar to daisyUI modal

packages/nextjs/app/builders/[address]/_components/UpdateLocationModal.tsx
```diff
@@ -0,0 +1,84 @@
...
+              <button
+                className="btn btn-outline"
+                disabled={isPending}
+                onClick={async () => {
+                  await updateLocation(null);
+                  setSelectedCountry("");
+                }}
+              >
+                Reset
+              </button>
```
Comment (technophile-04) [inline packages/nextjs/app/builders/[address]/_components/UpdateLocationModal.tsx:75]: In BGv3.5 we didn't had this reset button and was a bit unintuitive to reset/clear the location. So thought of having separate button only to clear the location

carletex:
> here is the discussion regarding it: https://github.com/drizzle-team/drizzle-orm/discussions/1104#discussioncomment-10814609

Let me know if this process goes smoothly!

packages/nextjs/services/database/config/postgresClient.ts
technophile-04, line 1: Actually, Vercel has removed the vecel/postrgress and now giving direct access to Neon checkout [this](https://neon.tech/docs/guides/vercel-postgres-transition-guide)


So this logic: 

```ts
const VERCEL_DB_STRING = "verceldb";
```

Acutally was not woking only, also if look at vercel env itnow has `neondb` instead of `verceldb` 

Also [`@vercel/postgress`](https://neon.tech/docs/guides/vercel-postgres-transition-guide#will-the-vercel-postgres-sdk-continue-to-work) won't be maintained. 

So that's why updated now this with `@neondatabse/serverless`  which is reccommded way. 

The reason for all this exploration and updating was lol I wanted to have batching update in my sync script and noticed the `@vercel/postgress` orm SDK didn't had it

packages/nextjs/scripts/database/index.ts
technophile-04, line 1: This is script is almost same as `postgrestClient.ts` the difference is just that, here we initialize `dotenv` (since in postgreasClient.ts it's initialized by Nextjs) and also export client with db `exports {client, db}` 

So that `script` using this can end the db connection after running.

packages/nextjs/scripts/database/sync_bg_locations.ts
```diff
@@ -0,0 +1,65 @@
...
+        updatedCount++;
+      }
+    }
+  }
+
+  if (db instanceof NeonHttpDatabase && updates.length > 0) {
+    await db.batch(updates as any);
+    updatedCount += updates.length;
+    console.log(`Batch updated ${updates.length} users.`);
+  }
```
technophile-04, line 52: This works nciely, tried attaching to personal neon account and : 
image

carletex:
Hey @technophile-04 , could you create a separate PR for the database changes (new serverless lib, neondb, etc).

I'm having some issues with #123 (if I wanted to use the existing Postgress client) and found some solutions, so I'll check with the new setup.

technophile-04:
Ok added a script called `sync_bg_locations.ts`.

I think we are gonna have some more similar sync script, so that's why created `scripts` dir and inside it we have `databse/sysnc_bg_location.ts` which does this syncing of location. To call this script you can `cd packages/nextjs && yarn sync:bg-locations`

The script uses `fetchAndCacheBGBuildersData` util which creates `bg_builders.json` file  fetching builders data from bg if not present inside `scripts/data` so that we don't burst bg backend again nd again. And this util can be consumed by other sync scripts too 🙌 

also has `scripts/databse/index.ts` which exposes `db` instance just for scipts so that those scripts can close the connection after they end.  

Happy to refactor things and hear more thoughts 🙌

carletex:
Thanks for the explanation Shiv!


I think this PR is too big, not only number of lines wise (which is not super big) but also functionality wise. It's changing the lib for the Database, adding the location functionality, and creating a new system for importing data from BG. This should probably be 3 PRs. Merging something like this is more likely to break production (and harder to review)

I think it could be:
1. Database changes
2. Location functionality.
3. Import

(This is what I tried to do with batches)

- I had the scripts in the services/database folder... seems to make more sense? We will have a bunch of import scripts from BG.
image

We can discuss in the PR/issue tho!

- I see that you are tweaking package.json to add the script. I think running it with `yarn tsx` should be enough, no? Maybe the issues I was having on https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/127#issuecomment-2809234145 was because of it haha (database wasn't avaiable):


Thanks !! Also, let's do a sync soon if you want ;)

technophile-04:
Removed the sync script setup  

Also created a branch in neon db with named `location` from `main`. And configured that URL to be used by vercel preview deployment and ran `yarn drizzle-kit migrate` on this `location` branch in db. So you could test it at 

https://speedrunethereum-v2-m2vd97nao-buidlguidldao.vercel.app/

--- 

I think before merging this: 

We need to run `yarn drizzle-kit migrate` for `staging` and `main` branch of neon DB 🙌

carletex:
Hey Shiv, thanks for this!

Pushed a couple of things:

1. https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/127/commits/e72abacc15cb4c06cca2ea6e2d9210c6155fe709 Because of the empty markup we were creating empty space when location / socials were not set. So returning `null` in those cases.

before:
image

after:
image

2. https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/127/commits/65b75dd07f278f628497cc01d5e781816dc66006

Changed the UX/UI a bit on the buttons
- I think CTA/Action should always be on the right, so switched the reset / update buttons
- Reset, just unselects the country (doesn't send the actual call, you need to click update location). For me, this is more natural, but let  me know what you think.
- reset => btn ghost to make it "less competitve"

Before:
image

After:
image

-----

> Also created a branch in neon db with named location from main. And configured that URL to be used by vercel preview deployment and ran yarn drizzle-kit migrate on this location branch in db.

Great stuff. I think when we merge, we should do a clean up (delete database branch + Vercel env var). Maybe this will be automated in the future.

> https://speedrunethereum-v2-m2vd97nao-buidlguidldao.vercel.app/

For preview branches, Vercel gives you two URL:
1. The one you pasted (specific to a given commit / deployment)
2. The general URL for the branch [speedrunethereum-v2-git-issue-97-buidlguidldao.vercel.app](https://speedrunethereum-v2-git-issue-97-buidlguidldao.vercel.app/)

Just mentioning in case people are trying to access to commit deployment URL and doesn't see it updated with the new commits.

> We need to run yarn drizzle-kit migrate for staging and main branch of neon DB 

Yes!

Note: for staging, you can also "reset from parent" in the Neon UI after you run the migrations on Main.

Approving this! If you agree with my changes, let's merge (and run the migrations), if not, let's discuss!

Thanks Shiv, great job!

carletex (approved):

carletex:
>  just pushed a very small nitpick so that we disable update btn if selectedCountry and existingLocation

Great! The little things <3

> ahh I thought "reset from parent" only worked when schema's were same

It basically recreates it again, so schema + data will match main

---

Looking great, merging!!

PS: Would love to talk about the import scripts (working on the batches one right now)

carletex:
(applied migration too)

PR #130, opened 2025-04-15
Protect drizzle config

carletex:
Just me being paranoid, but this is a easy tweak that can save our asses in some cases.

This will protect `yarn drizzle-kit` actions ( `migrate`, `push`). `seed` is already protected.

What do you guys think?


rin-st:
Good idea! 

But for now it doesnt work

```
yarn drizzle-kit migrate --force
Unrecognized options for command 'migrate': --force
```

`push` works, but just because it has `--force` [flag](https://orm.drizzle.team/docs/drizzle-kit-push#extended-list-of-configurations).

---

The best solution `cursor` proposed for me is to create the script
```
import { spawn } from "child_process";
import * as dotenv from "dotenv";
import * as path from "path";

dotenv.config({ path: path.resolve(__dirname, "../.env.development") });

const PRODUCTION_DATABASE_HOSTNAME = "cold-resonance";

if (process.env.POSTGRES_URL?.includes(PRODUCTION_DATABASE_HOSTNAME) && !process.argv.includes("--force")) {
  console.log("\n⚠️ You are pointing to the production database. Use `--force` if you want to continue.\n");
  process.exit(0);
}

// Remove the --force flag if it exists
const args = process.argv.slice(2).filter(arg => arg !== "--force");

// Spawn drizzle-kit with the remaining arguments
const drizzleKit = spawn("drizzle-kit", args, {
  stdio: "inherit",
  shell: process.platform === "win32",
});

drizzleKit.on("exit", code => {
  process.exit(code || 0);
});
```

and change `package.json`
image

Also, since `--force` conflicts with default `push` `--force` flag, it's better to change our flag (and hence all the `--force` in the script above) to something different, e.g `--force-update`

packages/nextjs/drizzle.config.ts
```diff
@@ -3,6 +3,13 @@ import { defineConfig } from "drizzle-kit";
 
 dotenv.config({ path: ".env.development" });
 
+const PRODUCTION_DATABASE_HOSTNAME = "cold-resonance";
+
+if (process.env.POSTGRES_URL?.includes(PRODUCTION_DATABASE_HOSTNAME) && !process.argv.includes("--force")) {
+  console.log("\n⚠️ You are pointing to the production database. Use `--force` if you want to continue.\n");
+  process.exit(0);
+}
```
technophile-04, line 11: For the problem which rinat mentioned what if instead of using the `--flag` we could set the env variable while running the command? 

eg: 

```
ALLOW_PROD_ACTIONS=true yarn drizzle-kit migrate
```

carletex: I've been thinking about this since Rinat's review.

This option was an improvment (since it simplifies everything) but it has something (as the --force one) that I don't like: people (me) sometime use CTRL+R to search for previous commands (e.g. push / migrate) and if you hit enter, (without noticing / ignore the ENV_VAR) you are f***

So went with another option. Pushing in 2 min

carletex:
> and we have to update .env.development to run migrate or any script to interact with neonDB too and lol I am always concerned committing and pushing that file to github by mistake and leaking the DB URL

Yes! Let's discuss this (here or in another place).

I see PROS/CONS. The think I liked about .env.dev changed on my git files, is that I know that I wasn't pointing to localhost. But yes, you can leak it.... so maybe we can have .env or .env.local? Now that we have PRODUCTION protected, I think we should go with a .gitignored file.

packages/nextjs/drizzle.config.ts
rin-st:
```
ALLOW_PROD_ACTIONS=true yarn drizzle-kit migrate
```

Yes, I've been thinking about that solution, but if someone will add `ALLOW_PROD_ACTIONS=true` to `.env`, than, again

> you are f***

rin-st:
I like the solution! 

I have some nitpicks which totally could be ignored

- it doesnt work with node 18. But well, db is still protected :D. 20+ works fine
image

- it takes only first symbol and then try to execute other symbols. Maybe in future we need to prevent that behavior. And also strict check of the answer to `y`/`Y`
image

- also we can make `yes` the right answer too but since it's only for us it's not needed for now

rin-st (approved):

carletex:
Thanks Rinat! Came with another solution (just tinkering, this PR is not critical / urgent)

- Works in node 18
- clean the ouput (doesn't execute the other symbols)

Create a util for it, since it might be useful for other scripts (imports, etc)

But let me know if this feels too overkill

rin-st:
> Works in node 18
> clean the ouput (doesn't execute the other symbols)

It's working for me too

> Create a util for it, since it might be useful for other scripts (imports, etc)

👍

rin-st (approved):

technophile-04 (approved):
Thanks @carletex!! Yup workinge nicely for me and nice approach like it! 

a small very nitpck and we can for sure ignore it not a big deal at all so just we can ingore but when choosing the default option (N) I have to hit enter twice. If I add letter n/N and hit enter it works nicely and also y/Y, but just I need to hit two enters selecting default but lol again not a big deal 

<details>

<summary>
    Demo video
</summary>


https://github.com/user-attachments/assets/b9f2a24f-4c45-4479-ba21-7b1bcea488c2



</details> 

---- 

> I see PROS/CONS. The think I liked about .env.dev changed on my git files, is that I know that I wasn't pointing to localhost. But yes, you can leak it.... so maybe we can have .env or .env.local? Now that we have PRODUCTION protected, I think we should go with a .gitignored file.

ohh yeah actually seeing it diff make sense! No strong opinions but maybe yeah since now have the check maybe we configure this drizzle config to use `.env.local`

packages/nextjs/drizzle.config.ts
```diff
@@ -1,8 +1,25 @@
...
 dotenv.config({ path: ".env.development" });
 
+const PRODUCTION_DATABASE_HOSTNAME = "cold-resonance";
+
+if (process.env.POSTGRES_URL?.includes(PRODUCTION_DATABASE_HOSTNAME)) {
+  process.stdout.write("\n⚠️ You are pointing to the production database. Are you sure you want to proceed? (y/N): ");
+
+  const result = spawnSync("tsx", [join(__dirname, "utils/prompt-confirm.ts")], {
+    stdio: "inherit",
+  });
```
Pabl0cks, line 15: Having an issue here on Windows, after being prompted, it would automatically show "Aborted" and kill the command (see sshot below). Adding this line it works great:

```
const result = spawnSync("tsx", [join(__dirname, "utils/prompt-confirm.ts")], {
    stdio: "inherit",
    shell: process.platform === "win32"
  });
```

Claude explanation:
> This is a known issue with Windows command shells (especially Git Bash) where the terminal's input mode and stdin inheritance to child processes don't work the same way as on Unix-like systems.
The simplest fix would be to add `shell: true` to the spawnSync options when on Windows:

<details><summary>Windows sshot</summary>

image

</details>

Pabl0cks (commented):
I love the security measure and the solution, GJ @carletex!!  🔥

Sorry for jumping in so late, you guys already came up with a great solution, and here comes the Windows guy to report issues 😅
I added a comment with a possible fix. Hopefully it doesn’t mess with the flow on Linux/Mac.

packages/nextjs/drizzle.config.ts
carletex: Since it seems like a quick fix I added it (works on Linux too). Can you check on macOS @rin-st @technophile-04 ?

PS: Does it work on WSL, Pablo?

carletex:
> a small very nitpck and we can for sure ignore it not a big deal at all so just we can ingore but when choosing the default option (N) I have to hit enter twice. If I add letter n/N and hit enter it works nicely and also y/Y, but just I need to hit two enters selecting default but lol again not a big deal

Noticed that too :sweat: Started to fix it but it was a bit tricky with the ascii representation of the New Line and differences on different OS, so I skipped it. I think we can leave with it tho :D

packages/nextjs/drizzle.config.ts
technophile-04: Its working great for me 🙌

technophile-04 (approved):

packages/nextjs/drizzle.config.ts
Pabl0cks: On WSL is working nice! 👌

carletex:
Thanks all!! Merging this. Will create another PR for the .env.dev / .env.local thing so we can discuss there.

PR #133, opened 2025-04-16
use `neondatabase/serverless` package for prod db instance

technophile-04:
### Description: 

Removed the vercel/postgress package and now using the `neondatabse/serverless` package 

as discussed in https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/127#issuecomment-2809234145

#### Research: 

> pasting below the comment from this [comment](https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/127#discussion_r2046655363) so we have details in this PR nicely 🙌

Actually, Vercel has removed the vecel/postrgress and now giving direct access to Neon checkout [this](https://neon.tech/docs/guides/vercel-postgres-transition-guide)

So this logic: 

```ts
const VERCEL_DB_STRING = "verceldb";
```

Acutally was not woking only, also if look at vercel env itnow has `neondb` instead of `verceldb` 

Also [`@vercel/postgress`](https://neon.tech/docs/guides/vercel-postgres-transition-guide#will-the-vercel-postgres-sdk-continue-to-work) won't be maintained. 

So that's why updated now this with `@neondatabse/serverless`  which is reccommded way. 

The reason for all this exploration and updating was lol I wanted to have batching update in my sync script and noticed the `@vercel/postgress` orm SDK didn't had it


carletex (approved):
After https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/133/commits/7890a753f559cb7bb46d76d1c6515c8ef4f9429a this works.

We found out that some actions (e.g. updateSocials) weren't working properly. The database was updated, but the page wasn't refresing (Even if you do hardrefresh). After debugging with Shiv, we found out that we need to do "pool". Not exactly sure how that works (Shiv is going to research a bit) but with the new Pool call, it works.

Merging!

technophile-04:
So the reason why it was not working with `http` client before and the reason we had to use `Pool` is: 

The `drizzle-orm/neon-http` (earlier way where updates were made but not reflecting) connects to Neon over HTTP, which is optimized for single, non-interactive transactions. 

But then we switched to `Pool` with `@neondatabase/serverless` and it uses WebSockets and is designed for interactive transactions or when you need session support. 

So like when we are using `Pool` we are actually using the websockets for connections and hence there was this reactive behaviour on website. 

So when to use `drizzle-orm/neon-http`, it's helpful to perform good optimized single transaction like `batch`. And also the best of using `drizzle` is that the abstractions with how they have created the library is if you use `drizzle-orm/neon-serverlss` `drizzle` it doesn't give you `batch` their on the `dbIstance` instead you have to use db Instance which is created via `drizzle-orm/neon-http` to perform batch.

carletex:
Thanks for the explanation, Shiv!

The only part that I don't understand is why the thing was still not showing after doing a hard refresh. Even if you use HTTP, you can refetch the data (Which we are doing) or do a page hard refresh (new HTTP connection) and the thing should show.

So there might be a layer of caching that I'm not aware of.

technophile-04:
#### Summary: 

> Maybe neon-http is more optimized for some stuff, but since this is only for scripts, using the regular `pg` driver should be more than enough

Ohh yess actually the way we did with current regular pg `Pool` driver is being used normally in postgress databse systems and looks good to me.

<details>

<summary>
    lol didn't want to fill this PR with huge wall of text : 
</summary>

#### Our current setup

We are using [`@neondatabase/serverless`](https://github.com/neondatabase/serverless?tab=readme-ov-file#neondatabaseserverless) (which is drop in replacement driver for `pg`). 

We are using `pg` Pool for connecting to local postgress database.

This both are basically postgress dirvers

Now, generally when we use `Pool` or `Client` exported from `pg` / `@neondatabse/serverless` it connects only via tcp / websockets.

But since neon databse also supports connection via HTTP, the `@neondatabse/serverless` package also export `neon(...)` function driver which allows querying over HTTP. 

> NOTE: Generally querying over HTTP is available in normal PG but it's a feature which Neon Database provides and hence the library also expose the `neon(...)` which allow us to do it.

And `drizzle-orm/neon-serverless`, `drizzle-or/neon-http` and `drizzle-orm/node-postgres` is basically a wrapper by drizzle which allows use this drivers and connect to database and give us some automcompletions etc.

And since neon allows two types of connections, drizzle also provide two types of wrapper `drizzle-orm/neon-serverless` and `drizzle-orm/neon-http`.
1. When using Pool from `@neondatabse/serverless` we use `drizzle-orm/neon-serverless`  (websocket connection)
2. When using neon()  `@neondatabse/serverless` we use `drizzle-orm/neon-http`  (http connection)

---

#### Transactions (interactive) [All PG DB support this]

Now this Drizzle ORM provides us with nice (interactive trasaction) `db.transaction` (transaction in PG dabatase is basically to group multiple SQL  statement into a single, all or nothing operation).

So when we do: 

```typescript
  await tx.update(...);
  await tx.update(...);
});
```

Drizzle gets "persistent", "interactive connection" (hence also requires websocket connection*) b/w DB and sends "multiple" SQL statements over this connection to databse.

1. Start transction (BEGIN;)
2. update row 
3. update row
4. end transction (COMMIT) Or (ROLLBACK) if any of in b/w things fails.

So the round trips here are more. I.e Client sending this SQL statments again and again. checkout the source [here](https://github.com/drizzle-team/drizzle-orm/issues/3828). 

This is very common in PG and trasaction are always used when we want to have atomic operation and multiple interactions with db.

This is what we did in this PR

> Oh, tried to run it on the `batches` database Neon branch and it seems that there's not tx support :(
> from https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/136#issuecomment-2819086141

Since for transactions (way how drizzle implements it as ORM) we need that persistent connection and hence we need to use `drizzle-orm/neon-serverless`(websocket) and hence while using `drizzle-orm/neon-http` (http conenction) it doesn't allow to do that. And hence we got the above error.  (kind of related to [this](https://www.answeroverflow.com/m/1149370348593217619)). 

#### Transaction (non-interactive)

These transactions are executed in a stateless manner, with all the transaction's operations and context included in a "single request" to the database. 

In drizzle we can can do this via [`db.batch()`](https://orm.drizzle.team/docs/batch-api) so when we do: 

```typescript
 db.batch([
   db.update(...),
   db.update(...)
 ]); 
```

Under the hood dirzzle constructs this whole as one SQL statement : 

```sql
BEGIN;
UPDATE table1 SET column_name1 = 'new_value' WHERE condition1;
UPDATE table2 SET column_name2 = 'new_value' WHERE condition2;
COMMIT;
 ```
 
source [here](https://github.com/drizzle-team/drizzle-orm/issues/3828)
 
 And this whole SQL statement is send in single HTTP request. Notice the difference in above and this method, in earlier method we first sent (BEGIN), then update, then (COMMIT or ROLLBACK). But Here we just send this whole SQL block  in once single request.

But since HTTP connection is not normal for common PG database the [`db.batch()`](https://orm.drizzle.team/docs/batch-api) api is only available for only certain databse like neon which supports HTTP.

And to use it we need to use `drizzle-orm/neon-http` instead of `drizzle-orm/neon-serverless`. 

So `drizzle-orm/neon-http` is usefull for this batchng stuff in single request.

---
#### Research summary: 


1. I think we should remove the `pg` library and use only `@neondatabse/serverless` for drivers since its drop-in replacement.
2. Since neondatabse has this extra feture of http, while configuring local databse instead of using normal postgreas docker, we should their neon-proxy local db setup which gives this methods as well as is more closer to prod neon databse.
   resource https://neon.tech/guides/local-development-with-neon#local-postgresql
3. While reading this all this blogs and setups, it seems that `drizzle-orm/neon-http` is suggested more in serverless environment, even in [drizzle TODO](https://orm.drizzle.team/docs/tutorials/drizzle-nextjs-neon) next app they use that instead of `drizzle-orm/neon-serverless`. The only blocker for not using `neon-http` is https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/133#issuecomment-2817971750 , but also as carlos mentioned there might be some cache layer, and it could be on NextJs side which was causing this.
4. So will play around more with `neon-http` for whole app, and if its seems good. Then we can only have it for scripts (batch) and for normal Next app and also local environment too since we will be using local postgres db too.

Lol tinkering on weekend, since this is not that top priority and just an optimization of stuff.

</details>

PR #134, opened 2025-04-17
use relative path for og image

technophile-04:
### Description: 

Use relative path for getting punks image


technophile-04:
I think #135 was good approach actually since this doesn't work

PR #136, opened 2025-04-17
Database lazy-loading (+ Batches import script)

carletex:
Started working on the scripts for importing Batches from BG and ended up going down the rabbit hole with Database instantiation. SO this PR might be more about discussing the patterns that the actual Batches script.

- Add bg-import scripts to `services/database/bg-imports`. Dated.

First, with only the script, I tried to run it like: `cd packages/nextjs` + `yarn tsx services/database/bg-imports/20250416_batch_data.ts`. It gave me an error :red_circle: 

The env vars weren't loading. This is because of how NodeJS handles import: it recursively loads all the imports, and since our `postgresClient` just directly loads the DB, it didn't have the proper data. This isn't a problem in the NextJS env (because Next does its thing while bootstraping the app) but it's an issue when using it externaly (like a script)

So, two solutions:

1. Just add a dotenv load in `postgresClient`. This is the quick & dirty fix. The thing that I don't like is that we are coupling the specific .env var to it....as we do in drizzle.config (in this case, it doesn't matter because it always runs in the cli, outside the Next env... but it's not the case on `postgresClient`)

2. If we want to "client" of `postgresClient` to load the env (nextjs, the script, etc) we need to do lazy loading:
- Dynamic imports on the client (we need to tweak the clients)
- Factory pattern, so we do db.init on the client (we need to tweak the clients)
- Proxy pattern (NO need to tweak the clients, the db gets init when is going to be used)

I've never used the Proxy pattern in JS, but went with that option. Everything seems to be working fine, and it only adds a bit of overload. It also makes sure that there is only one db instance (I think this was also happening before, because of how JS imports work). You can also "close" the connection when using the regular `Pool`

What do you guys think? Is it too much?

(P.S: The import script is just an example, and needs some tweaks... like upserts, checking if batch already exists, etc)

### DIFF
--- a/packages/nextjs/services/database/bg-imports/20250416_batch_data.ts
+++ b/packages/nextjs/services/database/bg-imports/20250416_batch_data.ts
@@ -1,7 +1,6 @@
 import { db } from "../config/postgresClient";
-import { users } from "../config/schema";
+import { batches, users } from "../config/schema";
 import { BatchStatus, BatchUserStatus } from "../config/types";
-import { createBatch } from "../repositories/batches";
 import * as dotenv from "dotenv";
 import { eq } from "drizzle-orm";
 import * as path from "path";
@@ -43,54 +42,76 @@ export async function importData() {
 
     console.log(`Found ${batchesData.length} batches and ${usersData.length} users belonging to a batch`);
 
-    // Save the mapping between Firebase ID and our database ID
-    const batchIdMap = new Map<string, number>();
+    // Start a transaction
+    await db.transaction(async tx => {
+      // Save the mapping between Firebase ID and our database ID
+      const batchIdMap = new Map<string, number>();
 
-    // Import batches
-    for (const batch of batchesData) {
-      const newBatch = {
-        name: `Batch ${batch.name}`,
-        startDate: new Date(batch.startDate),
-        status: batch.status.toLowerCase() as BatchStatus,
-        contractAddress: batch.contractAddress,
-        telegramLink: batch.telegramLink,
-      };
+      // Preload existing batches
+      const existingBatches = await tx.select().from(batches);
+      const existingBatchesByTelegramLink: Record<string, (typeof existingBatches)[0]> = {};
+      existingBatches.forEach(batch => {
+        if (batch.telegramLink) {
+          existingBatchesByTelegramLink[batch.telegramLink] = batch;
+        }
+      });
 
-      const createdBatch = await createBatch(newBatch);
-      console.log(`Imported batch: ${createdBatch.name} (ID: ${createdBatch.id})`);
+      // Import batches
+      for (const batch of batchesData) {
+        const newBatch = {
+          name: `Batch ${batch.name}`,
+          startDate: new Date(batch.startDate),
+          status: batch.status.toLowerCase() as BatchStatus,
+          contractAddress: batch.contractAddress,
+          telegramLink: batch.telegramLink,
+        };
 
-      // Store the mapping between Firebase ID and our database ID
-      batchIdMap.set(batch.name, createdBatch.id);
-    }
+        let createdBatch;
+        if (batch.telegramLink && batch.telegramLink in existingBatchesByTelegramLink) {
+          // Skip if the batch already exists in the database
+          createdBatch = existingBatchesByTelegramLink[batch.telegramLink];
+          console.log(`Found existing batch: ${createdBatch.name} (ID: ${createdBatch.id})`);
+        } else {
+          const result = await tx.insert(batches).values(newBatch).returning();
+          createdBatch = result[0];
+          console.log(`Imported batch: ${createdBatch.name} (ID: ${createdBatch.id})`);
+        }
 
-    // Update users with batch information
-    for (const user of usersData) {
-      const batchId = batchIdMap.get(user.batch.number);
-
-      if (!batchId) {
-        console.warn(`Batch ${user.batch.number} not found for user ${user.id}`);
-        continue;
+        // Store the mapping between Firebase ID and our database ID
+        batchIdMap.set(batch.name, createdBatch.id);
       }
 
-      const result = await db
-        .update(users)
-        .set({
-          batchId: batchId,
-          batchStatus: user.batch.status.toLowerCase() as BatchUserStatus,
-        })
-        .where(eq(users.userAddress, user.id));
+      // Update users with batch information
+      for (const user of usersData) {
+        const batchId = batchIdMap.get(user.batch.number);
 
-      if (result.rowCount === 0) {
-        console.warn(`User ${user.id} not found in database`);
-        continue;
+        if (!batchId) {
+          console.warn(`Batch ${user.batch.number} not found for user ${user.id}`);
+          continue;
+        }
+
+        // Use transaction for update
+        const result = await tx
+          .update(users)
+          .set({
+            batchId: batchId,
+            batchStatus: user.batch.status.toLowerCase() as BatchUserStatus,
+          })
+          .where(eq(users.userAddress, user.id));
+
+        if (result.rowCount === 0) {
+          console.warn(`User ${user.id} not found in database`);
+          continue;
+        }
+
+        console.log(`Updated user: ${user.id} with batch ${user.batch.number}`);
       }
-
-      console.log(`Updated user: ${user.id} with batch ${user.batch.number}`);
-    }
+    });
 
     console.log("Data import completed successfully!");
   } catch (error) {
     console.error("Error during data import:", error);
+    console.log("Transaction rolled back due to error.");
   } finally {
     await db.close();
     process.exit(0);
--- a/packages/nextjs/services/database/bg-imports/20250416_batch_data.ts
+++ b/packages/nextjs/services/database/bg-imports/20250416_batch_data.ts
@@ -1,0 +1,100 @@
+import { db } from "../config/postgresClient";
+import { users } from "../config/schema";
+import { BatchStatus, BatchUserStatus } from "../config/types";
+import { createBatch } from "../repositories/batches";
+import * as dotenv from "dotenv";
+import { eq } from "drizzle-orm";
+import * as path from "path";
+
+dotenv.config({ path: path.resolve(__dirname, "../../../.env.development") });
+
+const BG_BATCHES_ENDPOINT = `${process.env.NEXT_PUBLIC_BG_BACKEND}/batches`;
+const BG_USERS_WITH_BATCHES_ENDPOINT = `${process.env.NEXT_PUBLIC_BG_BACKEND}/builders/batches`;
+
+type BgBatch = {
+  id: string;
+  name: string;
+  status: string;
+  startDate: number;
+  telegramLink: string;
+  contractAddress: string;
+  totalParticipants: number;
+  graduates: number;
+};
+
+type BgBatchUser = {
+  id: string;
+  batch: {
+    number: string;
+    status: string;
+  };
+};
+
+export async function importData() {
+  try {
+    console.log("==== Importing BATCH DATA ====");
+    console.log("Fetching data from BG firebase...");
+
+    const batchesResponse = await fetch(BG_BATCHES_ENDPOINT);
+    const batchesData: BgBatch[] = await batchesResponse.json();
+
+    const usersResponse = await fetch(BG_USERS_WITH_BATCHES_ENDPOINT);
+    const usersData: BgBatchUser[] = await usersResponse.json();
+
+    console.log(`Found ${batchesData.length} batches and ${usersData.length} users belonging to a batch`);
+
+    // Save the mapping between Firebase ID and our database ID
+    const batchIdMap = new Map<string, number>();
+
+    // Import batches
+    for (const batch of batchesData) {
+      const newBatch = {
+        name: `Batch ${batch.name}`,
+        startDate: new Date(batch.startDate),
+        status: batch.status.toLowerCase() as BatchStatus,
+        contractAddress: batch.contractAddress,
+        telegramLink: batch.telegramLink,
+      };
+
+      const createdBatch = await createBatch(newBatch);
+      console.log(`Imported batch: ${createdBatch.name} (ID: ${createdBatch.id})`);
+
+      // Store the mapping between Firebase ID and our database ID
+      batchIdMap.set(batch.name, createdBatch.id);
+    }
+
+    // Update users with batch information
+    for (const user of usersData) {
+      const batchId = batchIdMap.get(user.batch.number);
+
+      if (!batchId) {
+        console.warn(`Batch ${user.batch.number} not found for user ${user.id}`);
+        continue;
+      }
+
+      const result = await db
+        .update(users)
+        .set({
+          batchId: batchId,
+          batchStatus: user.batch.status.toLowerCase() as BatchUserStatus,
+        })
+        .where(eq(users.userAddress, user.id));
+
+      if (result.rowCount === 0) {
+        console.warn(`User ${user.id} not found in database`);
+        continue;
+      }
+
+      console.log(`Updated user: ${user.id} with batch ${user.batch.number}`);
+    }
+
+    console.log("Data import completed successfully!");
+  } catch (error) {
+    console.error("Error during data import:", error);
+  } finally {
+    await db.close();
+    process.exit(0);
+  }
+}
+
+importData();
--- a/packages/nextjs/services/database/bg-imports/20250416_batch_data.ts
+++ b/packages/nextjs/services/database/bg-imports/20250416_batch_data.ts
@@ -68,9 +68,11 @@ export async function importData() {
 
         let createdBatch;
         if (batch.telegramLink && batch.telegramLink in existingBatchesByTelegramLink) {
-          // Skip if the batch already exists in the database
-          createdBatch = existingBatchesByTelegramLink[batch.telegramLink];
-          console.log(`Found existing batch: ${createdBatch.name} (ID: ${createdBatch.id})`);
+          // Update existing batch with any new information
+          const existingBatch = existingBatchesByTelegramLink[batch.telegramLink];
+          const result = await tx.update(batches).set(newBatch).where(eq(batches.id, existingBatch.id)).returning();
+          createdBatch = result[0];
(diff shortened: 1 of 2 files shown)


technophile-04:
Been researching about this and how to handle it. I have been looking at some of the popular full stack templates with drizzle, neon and Next and how they handle this flow. Also some youtube videos too and just found couple of decent example not that good enough and the had this: 

> Just add a dotenv load in postgresClient. This is the quick & dirty fix. The thing that I don't like is that we are coupling the specific .env var to it....as we do in drizzle.config (in this case, it doesn't matter because it always runs in the cli, outside the Next env... but it's not the case on postgresClient)

rin-st:
Agree, great solution! Played with it a bit and it works well!

packages/nextjs/services/database/config/postgresClient.ts
```diff
@@ -4,26 +4,66 @@ import { drizzle as drizzleNeon } from "drizzle-orm/neon-serverless";
...
+    dbInstance = null;
+  }
 }
 
-export { db };
+// Create a proxy to intercept all property accesses and method calls
+const dbProxy = new Proxy(
+  {},
+  {
+    get: (target, prop) => {
```
rin-st, line 54: nit: we can add type safety here: `prop: keyof DbProxy` and remove `any` at line 60

packages/nextjs/services/database/config/postgresClient.ts
```diff
@@ -4,26 +4,66 @@ import { drizzle as drizzleNeon } from "drizzle-orm/neon-serverless";
...
+      return (db as any)[prop];
+    },
+  },
+);
+
+type DbProxy = DbInstance & {
+  close: () => Promise<void>;
+};
+
+export const db = dbProxy as unknown as DbProxy;
```
rin-st, line 69: nit: works for me without unknown

carletex:
Thanks for the feedback!!

> Ohh regarding having the batch operation i.e helpful withing scripts, we need create neon db instance via drizzle-orm/neon-http and dbInstance current instance won't work and doesn't have batch operation. Checkout https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/133#issuecomment-2817742571 for more info 🙌

This made me think: we probably want batch operations for all our import scripts, right? Seems to be the safer (all or nothing import) and efficient way (many statements for tx)

=> To make that happen, we need `drizzle-orm/neon-http`

So there are two options:
1. Instantiate the DB on each import script using `neon-http`.
  - CONS: Repeating code (instantiating database) + we can't use repositories. Not a big deal I think.
2. Tweak PosgresClient to use `neon-http` in some case (detect if script / nextjs enviroments or env var or whatever)
  - CONS: Making it more complex

If we go with 1, I guess this PR won't make sense.

A quick and dirty example of 2:

```tsx
// Use something like this to detect if script?
const isScript = !!process.argv[1]?.includes("scripts/");
const isScript = !process.env.NEXT_RUNTIME;
```

```tsx
if (process.env.POSTGRES_URL?.includes(NEON_DB_STRING)) {
    if (isScript) {
    // Instanciate using neon-http
    } else {
    const neonPool = new NeonPool({ connectionString: process.env.POSTGRES_URL as string });
    dbInstance = drizzleNeon(neonPool, { schema, casing: "snake_case" });
   } 
}
```

Thoughts?

technophile-04:
Umm not sure if 2 seems too magical. And in the future if we want to use `http` client in dir other than `scripts` we need to add that condition in `postgreasClient.ts` 

But also for option 1: 

> CONS: Repeating code (instantiating database) + we can't use repositories. Not a big deal I think

Like here we plan to initiate db in each script file then we also need to replicate the whole thing from `postgresClient.ts` since we have to do `if branching`  when postgres URL is local db we want normal `pg` pool client to make request instead of neonHttp db. 

Maybe we create a `postgreatScriptClient.ts` which always returns HTTP client if `neondb` string is present and normal pg client if it's localhost? 


> If we go with 1, I guess this PR won't make sense.

Umm I feel more confident of having this code (instead of js handling initialization stuff and caching the object it self with imports), here seems like we do more verbose dbInstance cache.

But yeah no strong opinion, even if we keep it as it as `main` branch.  

Also yup it's totally fine to repeat code maybe since this scripts will be very rarely touched and are like almost one time only.

packages/nextjs/services/database/config/postgresClient.ts
```diff
@@ -4,26 +4,66 @@ import { drizzle as drizzleNeon } from "drizzle-orm/neon-serverless";
...
+let poolInstance: Pool | null = null;
+
+function getDb(): DbInstance {
+  if (dbInstance) {
+    return dbInstance;
+  }
+
+  const NEON_DB_STRING = "neondb";
+  if (process.env.POSTGRES_URL?.includes(NEON_DB_STRING)) {
+    const neonPool = new NeonPool({ connectionString: process.env.POSTGRES_URL as string });
```
technophile-04, line 19: here instead of assining to new variable `neonPool` we assign it to `poolInstance` so that we call end on neon pool instance?

carletex:
> Umm not sure if 2 seems too magical. And in the future if we want to use http client in dir other than scripts we need to add that condition in postgreasClient.ts

I think the second option (not sure if it works, or similar) is less magical: `const isScript = !process.env.NEXT_RUNTIME;`
We are basically detecting:
- Is Neon database?
- Are we on Next JS runtime?

If so => use Neon pool

> Maybe we create a postgreatScriptClient.ts which always returns HTTP client if neondb string is present and normal pg client if it's localhost?

Yes, this makes a lot sense too (naming TBD :D). It's basically the same as the other option but more explicit, less magical... but we are also duplicating 99% of the code.

No strong opinion either, both are fine and are not a big deal.

> Umm I feel more confident of having this code (instead of js handling initialization stuff and caching the object it self with imports), here seems like we do more verbose dbInstance cache.

Oooh, now you are sold on the Proxy thing haha. Yes, ok, happy to keep this.

rin-st:
> 2. Tweak PosgresClient to use `neon-http` in some case (detect if script / nextjs enviroments or env var or whatever)
> 
> * CONS: Making it more complex
> 
> If we go with 1, I guess this PR won't make sense.
> 
> A quick and dirty example of 2:
> 
> ```tsx
> // Use something like this to detect if script?
> const isScript = !!process.argv[1]?.includes("scripts/");
> const isScript = !process.env.NEXT_RUNTIME;
> ```
> 
> ```tsx
> if (process.env.POSTGRES_URL?.includes(NEON_DB_STRING)) {
>     if (isScript) {
>     // Instanciate using neon-http
>     } else {
>     const neonPool = new NeonPool({ connectionString: process.env.POSTGRES_URL as string });
>     dbInstance = drizzleNeon(neonPool, { schema, casing: "snake_case" });
>    } 
> }
> ```
> 
> Thoughts?

No strong opinion too, but for me the second option looks better

packages/nextjs/services/database/config/postgresClient.ts
carletex: I think it makes sense to do that for consistency. 

f2eefde671c0

But, when would you use close in a neon pool? I guess since this is the serverless one, it will never happen, right?

packages/nextjs/services/database/config/postgresClient.ts
carletex: Nice, thanks! e25467bb222f

packages/nextjs/services/database/config/postgresClient.ts
carletex: Oops. d7dcb58c32f1

carletex:
> No strong opinion too, but for me the second option looks better

Great, thanks!

Shiv also agrees that we want to do 2. But we have now options (a) or (b) (see previous comments)

a) more logic in postgresClient.ts
b) postgresScriptClient.ts

What do you think, @rin-st ?

carletex:
Well, maybe Shiv wanted 1 (but with a twist: using a file instead of replicating on every script)

technophile-04:
> The thing I like the most about this option is that we have a single source of thruth (even if has some weird logic)
>  is less magical: const isScript = !process.env.NEXT_RUNTIME; 

Ohh my bad yess completely missed the second `isScript`! Yes yes options 2 sounds niceee!! Since we will be only using the code and it's very deterministic only with this soo all good 🙌

carletex:
updated `postgresClient` in https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/136/commits/aa1512191f850cd0f15abca4c721581d6035f3d2 to use `NEXT_RUNTIME`

It's not super well documented, but the variable is there https://nextjs.org/docs/app/building-your-application/optimizing/instrumentation#importing-runtime-specific-code. The important thing is that it will be undefined when running scripts (or in non-NEXTJS runtimes)

I think this is clean enough. The postgres connection lives in a single source of truth, and it can be used in NextJS & scripts runtime + local & Neon database.

---

Also updated the import script for the batches: https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/136/commits/be22eade74324714755ea1f03bde6c715680e782 e1f7f4f7afe8

- Uses batched txs. Since the data is small, I'm doing everything on a single chunk. The good thing is that it would revert if anything fails. But we should use chunks if the data is large
- Checking if the batch already exists, so we can run this script anytime we want to sync with BG3.5. I'm updating (instead of skipping) in case some prop changes (like status). I'm using the `telegramLink` as the reference, I don't think that changes... :D

Also happy to split this into two PRs if needed.

carletex:
Oh, tried to run it on the `batches` database Neon branch :(

image

carletex:
Update: If I use the regular `pg` Pool, it works great (as in localhost), So i pushed  https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/136/commits/5cc5c1535361dbc4ee2b7fea1587709a0d92792b

Maybe neon-http is more optimized for some stuff, but since this is only for script, using the regular `pg` driver should be more than enough

Or am I missing something with neon-http and the batched tx? cc @technophile-04

technophile-04:
#### Summary: 

> Maybe neon-http is more optimized for some stuff, but since this is only for scripts, using the regular `pg` driver should be more than enough

Ohh yess actually the way we did with current regular pg `Pool` driver is being used normally in postgress databse systems and looks good to me.

<details>

<summary>
    lol didn't want to fill this PR with huge wall of text : 
</summary>

#### Our current setup

We are using [`@neondatabase/serverless`](https://github.com/neondatabase/serverless?tab=readme-ov-file#neondatabaseserverless) (which is drop in replacement driver for `pg`). 

We are using `pg` Pool for connecting to local postgress database.

This both are basically postgress dirvers

Now, generally when we use `Pool` or `Client` exported from `pg` / `@neondatabse/serverless` it connects only via tcp / websockets.

But since neon databse also supports connection via HTTP, the `@neondatabse/serverless` package also export `neon(...)` function driver which allows querying over HTTP. 

> NOTE: Generally querying over HTTP is **NOT** available in normal PG db's but it's a feature which Neon Database provides and hence the library also expose the `neon(...)` which allow us to do it.

And `drizzle-orm/neon-serverless`, `drizzle-or/neon-http` and `drizzle-orm/node-postgres` is basically a wrapper by drizzle which allows use this drivers and connect to database and give us some automcompletions etc.

And since neon allows two types of connections, drizzle also provide two types of wrapper `drizzle-orm/neon-serverless` and `drizzle-orm/neon-http`.
1. When using Pool from `@neondatabse/serverless` we use `drizzle-orm/neon-serverless`  (websocket connection)
2. When using neon()  `@neondatabse/serverless` we use `drizzle-orm/neon-http`  (http connection)

---

#### Transactions (interactive) [All PG DB support this]

Now this Drizzle ORM provides us with nice (interactive trasaction) `db.transaction` (transaction in PG dabatase is basically to group multiple SQL  statement into a single, all or nothing operation).

So when we do: 

```typescript
  await tx.update(...);
  await tx.update(...);
});
```

Drizzle gets "persistent", "interactive connection" (hence also requires websocket connection*) b/w DB and sends "multiple" SQL statements over this connection to databse.

1. Start transction (BEGIN;)
2. update row 
3. update row
4. end transction (COMMIT) Or (ROLLBACK) if any of in b/w things fails.

So the round trips here are more. I.e Client sending this SQL statments again and again. checkout the source [here](https://github.com/drizzle-team/drizzle-orm/issues/3828). 

This is very common in PG and trasaction are always used when we want to have atomic operation and multiple interactions with db.

This is what we did in this PR

> Oh, tried to run it on the `batches` database Neon branch and it seems that there's not tx support :(
> from https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/136#issuecomment-2819086141

Since for transactions (way how drizzle implements it as ORM) we need that persistent connection and hence we need to use `drizzle-orm/neon-serverless`(websocket) and hence while using `drizzle-orm/neon-http` (http conenction) it doesn't allow to do that. And hence we got the above error.  (kind of related to [this](https://www.answeroverflow.com/m/1149370348593217619)). 

#### Transaction (non-interactive)

These transactions are executed in a stateless manner, with all the transaction's operations and context included in a "single request" to the database. 

In drizzle we can can do this via [`db.batch()`](https://orm.drizzle.team/docs/batch-api) so when we do: 

```typescript
 db.batch([
   db.update(...),
   db.update(...)
 ]); 
```

Under the hood dirzzle constructs this whole as one SQL statement : 

```sql
BEGIN;
UPDATE table1 SET column_name1 = 'new_value' WHERE condition1;
UPDATE table2 SET column_name2 = 'new_value' WHERE condition2;
COMMIT;
 ```
 
source [here](https://github.com/drizzle-team/drizzle-orm/issues/3828)
 
 And this whole SQL statement is send in single HTTP request. Notice the difference in above and this method, in earlier method we first sent (BEGIN), then update, then (COMMIT or ROLLBACK). But Here we just send this whole SQL block  in once single request.

But since HTTP connection is not normal for common PG database the [`db.batch()`](https://orm.drizzle.team/docs/batch-api) api is only available for only certain databse like neon which supports HTTP.

And to use it we need to use `drizzle-orm/neon-http` instead of `drizzle-orm/neon-serverless`. 

So `drizzle-orm/neon-http` is usefull for this batchng stuff in single request.

---
#### Research summary: 


1. I think we should remove the `pg` library and use only `@neondatabse/serverless` for drivers since its drop-in replacement.
2. Since neondatabse has this extra feture of http, while configuring local databse instead of using normal postgreas docker, we should their neon-proxy local db setup which gives this methods as well as is more closer to prod neon databse.
   resource https://neon.tech/guides/local-development-with-neon#local-postgresql
3. While reading this all this blogs and setups, it seems that `drizzle-orm/neon-http` is suggested more in serverless environment, even in [drizzle TODO](https://orm.drizzle.team/docs/tutorials/drizzle-nextjs-neon) next app they use that instead of `drizzle-orm/neon-serverless`. The only blocker for not using `neon-http` is https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/133#issuecomment-2817971750 , but also as carlos mentioned there might be some cache layer, and it could be on NextJs side which was causing this.
4. So will play around more with `neon-http` for whole app, and if its seems good. Then we can only have it for scripts (batch) and for normal Next app and also local environment too since we will be using local postgres db too.

Lol tinkering on weekend, since this is not that top priority and just an optimization of stuff.

</details>

File: packages/nextjs/services/database/config/postgresClient.ts (hunk 3 above)
Comment (technophile-04) [reply packages/nextjs/services/database/config/postgresClient.ts]: Ohh Ohh true!!

File: packages/nextjs/services/database/bg-imports/20250416_batch_data.ts [hunk 4]
```diff
@@ -0,0 +1,123 @@
...
+
+    const usersResponse = await fetch(BG_USERS_WITH_BATCHES_ENDPOINT);
+    const usersData: BgBatchUser[] = await usersResponse.json();
+
+    console.log(`Found ${batchesData.length} batches and ${usersData.length} users belonging to a batch`);
+
+    // Start a transaction
+    await db.transaction(async tx => {
+      // Save the mapping between Firebase ID and our database ID
+      const batchIdMap = new Map<string, number>();
```
rin-st, line 48: change `number` to `string`

packages/nextjs/services/database/bg-imports/20250416_batch_data.ts
```diff
@@ -0,0 +1,123 @@
...
+        // Store the mapping between Firebase ID and our database ID
+        batchIdMap.set(batch.name, createdBatch.id);
+      }
+
+      // Update users with batch information
+      for (const user of usersData) {
+        const batchId = batchIdMap.get(user.batch.number);
+
+        if (!batchId) {
+          console.warn(`Batch ${user.batch.number} not found for user ${user.id}`);
```
rin-st, line 91: `Batch for user ${user.id} not found`

packages/nextjs/services/database/bg-imports/20250416_batch_data.ts
```diff
@@ -0,0 +1,123 @@
...
+            batchStatus: user.batch.status.toLowerCase() as BatchUserStatus,
+          })
+          .where(eq(users.userAddress, user.id));
+
+        if (result.rowCount === 0) {
+          console.warn(`User ${user.id} not found in database`);
+          continue;
+        }
+
+        console.log(`Updated user: ${user.id} with batch ${user.batch.number}`);
```
rin-st, line 109: `Updated user: ${user.id} with batch ID ${batchId}`

technophile-04 (approved):
Tysm @carletex for all the tinkering and my bad sorry, I should have done some tinkering before putting the theoretical explanation 

Just tried playing with the branch and yeah things seem broken :(

> So not sure, having this PR as it is might make more sense.... since it works nicely for both Next runtime + scripts runtime.  

Ohh yes yess! Love this changes and everything works nicely too and also it is actually optimised and very good! 

I feel we should keep things as it since this is also a standard and nice way of doing things 🙌 We can always overoptimize stuff with `neon-http` once we hit bottle neck, which I don't think we will hit at all only. 

We should merge this! Thanks soo much for tinkering and PR!

carletex:
lool no need to apologize, man. You did a great job and thanks to you I undertood a bunch of things.

I think the current setup works nicely. We can rethink it at some point or if something doesn't work properly.

Merging!

carletex:
PS: Ran the import script on `staging`. We can run it on `production` whenever we want, I think we can wait for now.

PS2: Had the issue that Shiv mentioned: some users are not present in SRE (but they are in BG3.5). Will create an issue

Issue #138, opened 2025-04-20
Builds schema and seed


Pabl0cks:
Add builds to the schema and add seed data examples.

<details><summary>Firebase builds data structure</summary>

![Image](https://github.com/user-attachments/assets/1451657f-42c5-4e09-a8f8-7179006d7821)

</details>

**Some database decisions to make:**

- Store builder + coBuilders or store a single builders field

- Store likes?

- Builds categorization

Issue #140, opened 2025-04-20
Import builds from BG with data curation


Pabl0cks:

carletex:
#162

PR #142, opened 2025-04-21
Builds schema and data seed

Pabl0cks:
Creating this draft to discuss the way to tackle #138

- Initial database fields similar to BG builds. Some notes:
   - Changed `id` to autonumeric field
   - Didn't add `featured` field, thinking we probably should be neutral in SRE and don't feature builds by our own
   - Importing "BG like system" to have a initial social validation system that we could use to sort builds.
   - Importing the `builderAddress` + `coBuilderAddresses` approach because I like the idea of having a main builder (owner?) responsible of having up to date the build data, which is a SRE builder, and coBuilders could be builders not registered in SRE.
   - Initially only using one categorized field as we had in BG (build type), but I think we should probably improve this categorization by adding 1 or 2 extra fields.

**Categorization proposal idea** (with examples)

#### **Level 1: Build Type (high level categorization similar to BG)**
- Dapp
- Extension
- Infrastructure
- Challenge submission
- Education
- Design
- Other

#### **Level 2: Build Category**
- DeFi
- Game
- NFT
- Governance
- Social
- Marketplace
- Dev tooling

#### **Level 3: Tech Stack**
- Scaffold-ETH / Scaffold-ETH-2
- Hardhat
- Foundry
- Next.js
- Ethers
- Viem
- Wagmi

In case we decide to go to a multiple field categorization =>  TBD how to manage the different categorizations in database.  Should all of them be enums, or just build type - level 1?

### DIFF
--- a/packages/nextjs/services/database/config/schema.ts
+++ b/packages/nextjs/services/database/config/schema.ts
@@ -104,12 +104,54 @@ export const userChallenges = pgTable(
   ],
 );
 
+export const builds = pgTable(
+  "builds",
+  {
+    id: serial().primaryKey(),
+    name: varchar({ length: 255 }).notNull(),
+    desc: text(),
+    buildType: varchar({ length: 50 }),
+    builderAddress: varchar({ length: 42 })
+      .notNull()
+      .references(() => users.userAddress),
+    coBuilderAddresses: varchar({ length: 42 }).array(),
+    demoUrl: varchar({ length: 255 }),
+    videoUrl: varchar({ length: 255 }),
+    imageUrl: varchar({ length: 255 }),
+    githubUrl: varchar({ length: 255 }),
+    submittedTimestamp: timestamp().notNull().defaultNow(),
+  },
+  table => [
+    index("build_builder_idx").on(table.builderAddress),
+    // Regular index for now - if we see performance issues with array queries later,
+    // we can explore adding a GIN index, would need to verify Neon/Drizzle support for GIN indexes before implementing
+    index("build_co_builders_idx").on(table.coBuilderAddresses),
+  ],
+);
+
+export const buildLikes = pgTable(
+  "build_likes",
+  {
+    id: serial().primaryKey(),
+    buildId: integer()
+      .notNull()
+      .references(() => builds.id),
+    likerAddress: varchar({ length: 42 })
+      .notNull()
+      .references(() => users.userAddress),
+    likedAt: timestamp().notNull().defaultNow(),
+  },
+  table => [uniqueIndex("build_like_unique_idx").on(table.buildId, table.likerAddress)],
+);
+
 export const usersRelations = relations(users, ({ many, one }) => ({
   userChallenges: many(userChallenges),
   batch: one(batches, {
     fields: [users.batchId],
     references: [batches.id],
   }),
+  builds: many(builds),
+  buildLikes: many(buildLikes),
 }));
 
 export const challengesRelations = relations(challenges, ({ many }) => ({
@@ -129,4 +171,23 @@ export const userChallengesRelations = relations(userChallenges, ({ one }) => ({
 
 export const batchesRelations = relations(batches, ({ many }) => ({
   users: many(users),
+}));
+
+export const buildsRelations = relations(builds, ({ one, many }) => ({
+  builder: one(users, {
+    fields: [builds.builderAddress],
+    references: [users.userAddress],
+  }),
+  likes: many(buildLikes),
+}));
+
+export const buildLikesRelations = relations(buildLikes, ({ one }) => ({
+  build: one(builds, {
+    fields: [buildLikes.buildId],
+    references: [builds.id],
+  }),
+  liker: one(users, {
+    fields: [buildLikes.likerAddress],
+    references: [users.userAddress],
+  }),
 }));
--- a/packages/nextjs/services/database/migrations/meta/0003_snapshot.json
+++ b/packages/nextjs/services/database/migrations/meta/0003_snapshot.json
@@ -1,0 +1,656 @@
+{
+  "id": "d12f6801-a2f6-4355-9da4-807f7289de70",
+  "prevId": "9c5331d3-f890-49c2-9c37-cd28d054ab7e",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.batches": {
+      "name": "batches",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "serial",
+          "primaryKey": true,
+          "notNull": true
+        },
+        "name": {
+          "name": "name",
+          "type": "varchar(255)",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "start_date": {
+          "name": "start_date",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "status": {
+          "name": "status",
+          "type": "batch_status_enum",
+          "typeSchema": "public",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "contract_address": {
+          "name": "contract_address",
+          "type": "varchar(42)",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "telegram_link": {
+          "name": "telegram_link",
+          "type": "varchar(255)",
+          "primaryKey": false,
+          "notNull": false
+        }
+      },
+      "indexes": {},
+      "foreignKeys": {},
+      "compositePrimaryKeys": {},
+      "uniqueConstraints": {},
+      "policies": {},
+      "checkConstraints": {},
+      "isRLSEnabled": false
+    },
+    "public.build_likes": {
+      "name": "build_likes",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "serial",
+          "primaryKey": true,
+          "notNull": true
+        },
+        "build_id": {
+          "name": "build_id",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "liker_address": {
+          "name": "liker_address",
+          "type": "varchar(42)",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "liked_at": {
+          "name": "liked_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "now()"
+        }
+      },
+      "indexes": {
+        "build_like_unique_idx": {
+          "name": "build_like_unique_idx",
+          "columns": [
+            {
+              "expression": "build_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            },
+            {
+              "expression": "liker_address",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": true,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        }
+      },
+      "foreignKeys": {
+        "build_likes_build_id_builds_id_fk": {
+          "name": "build_likes_build_id_builds_id_fk",
+          "tableFrom": "build_likes",
+          "tableTo": "builds",
+          "columnsFrom": [
+            "build_id"
+          ],
+          "columnsTo": [
+            "id"
+          ],
+          "onDelete": "no action",
+          "onUpdate": "no action"
+        },
+        "build_likes_liker_address_users_user_address_fk": {
+          "name": "build_likes_liker_address_users_user_address_fk",
+          "tableFrom": "build_likes",
+          "tableTo": "users",
+          "columnsFrom": [
+            "liker_address"
+          ],
+          "columnsTo": [
+            "user_address"
+          ],
+          "onDelete": "no action",
+          "onUpdate": "no action"
+        }
+      },
+      "compositePrimaryKeys": {},
+      "uniqueConstraints": {},
+      "policies": {},
+      "checkConstraints": {},
+      "isRLSEnabled": false
+    },
+    "public.builds": {
+      "name": "builds",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "serial",
+          "primaryKey": true,
+          "notNull": true
+        },
+        "name": {
+          "name": "name",
+          "type": "varchar(255)",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "desc": {
+          "name": "desc",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "build_type": {
+          "name": "build_type",
+          "type": "varchar(50)",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "builder_address": {
+          "name": "builder_address",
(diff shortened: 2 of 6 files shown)


Pabl0cks:
After speaking with Carlos, did a few changes:

- Base branch => `main`
- Build id is now a `varchar({ length: 40 })` instead of `text` (36 from uuid + 4 buffer for edge cases)
- Update migration to resolve conflicts with `main`

packages/nextjs/services/database/config/schema.ts
```diff
@@ -105,12 +99,71 @@ export const userChallenges = pgTable(
   ],
 );
 
+export const builds = pgTable(
+  "builds",
+  {
+    // Using varchar type for legacy imported IDs, with a default of gen_random_uuid() for auto-generation if not specified
+    id: varchar({ length: 40 })
+      .primaryKey()
+      .default(sql`gen_random_uuid()::text`),
```
carletex, line 108: I've been thinking about this, and I think we can implement a better solution. `uuid` field has several advantages like optimization, better indexing, etc... so we should try to keep it.

So we could use a deterministic function that converts firebaseID to uuids (on the import script). Something like:

```ts
import { v5 as uuidv5 } from "uuid";

 // RANDOM UUID to server as a base. It's safe for this to be public
const FIREBASE_TO_UUID_NAMESPACE = "ddeb27fb-d9a0-4624-be4d-4615062daed4";

// This will always return the same uuid with a given firebase ID
export function convertFirebaseIdToUuid(firebaseId: string) {
  return uuidv5(firebaseId, NAMESPACE);
}
```

If we do redirections on BG3.5, we can always compute this an redirect to the right uuid.

What do you think?

packages/nextjs/services/database/config/schema.ts
```diff
@@ -19,19 +20,12 @@ export function lower(address: AnyPgColumn): SQL {
...
-export const batchUserStatusEnum = pgEnum("batch_user_status_enum", [
-  BatchUserStatus.GRADUATE,
-  BatchUserStatus.CANDIDATE,
-]);
+export const reviewActionEnum = pgEnum("review_action_enum", ReviewAction);
+export const userRoleEnum = pgEnum("user_role_enum", UserRole);
+export const batchStatusEnum = pgEnum("batch_status_enum", BatchStatus);
+export const batchUserStatusEnum = pgEnum("batch_user_status_enum", BatchUserStatus);
+export const buildTypeEnum = pgEnum("build_type_enum", BuildType);
+export const buildCategoryEnum = pgEnum("build_category_enum", BuildCategory);
```
carletex, line 28: What do you think about this @technophile-04 ? And CONS?

It's less verbose / explicit... but it make sense? I guess the enum returns an array?

packages/nextjs/services/database/config/schema.ts
```diff
@@ -105,12 +99,71 @@ export const userChallenges = pgTable(
...
+    demoUrl: varchar({ length: 255 }),
+    videoUrl: varchar({ length: 255 }),
+    imageUrl: varchar({ length: 255 }),
+    githubUrl: varchar({ length: 255 }),
+    submittedTimestamp: timestamp().notNull().defaultNow(),
+  },
+  table => [
+    index("build_name_idx").on(table.name),
+    index("build_type_idx").on(table.buildType),
+    index("build_category_idx").on(table.buildCategory),
```
carletex, line 122: question: Are we sure that we'd need these indexes?

Not sure, especially about the "name" one. Wondering if we should use `tsvector` for full search (if we really want to use it)

Maybe we could start only by buildtype / category and review on the future when thinking about name / description search.

packages/nextjs/services/database/config/schema.ts
```diff
@@ -105,12 +99,71 @@ export const userChallenges = pgTable(
...
+      .notNull()
+      .references(() => builds.id),
+    userAddress: varchar({ length: 42 })
+      .notNull()
+      .references(() => users.userAddress),
+    isOwner: boolean().default(false).notNull(),
+  },
+  table => [
+    primaryKey({ columns: [table.buildId, table.userAddress] }),
+    index("build_builder_build_idx").on(table.buildId),
```
carletex, line 139: Do we need this `build_builder_build_idx` index? I don't understand it fully, but it seems that the composite pk will cover it (fetch all builds from a single builders)

(will love to learn more!)

packages/nextjs/services/database/config/schema.ts
```diff
@@ -105,12 +99,71 @@ export const userChallenges = pgTable(
...
+    index("build_builder_user_idx").on(table.userAddress),
+  ],
+);
+
+export const buildLikes = pgTable(
```
carletex, line 144: Not a big deal, but I think we talked about renaming this to "stars"?

packages/nextjs/services/database/config/schema.ts
```diff
@@ -105,12 +99,71 @@ export const userChallenges = pgTable(
...
+  "build_likes",
+  {
+    id: serial().primaryKey(),
+    buildId: varchar({ length: 40 })
+      .notNull()
+      .references(() => builds.id),
+    likerAddress: varchar({ length: 42 })
```
carletex, line 151: I think this could be `userAddress` not liker :D

packages/nextjs/services/database/migrations/meta/0003_snapshot.json
```diff
@@ -447,4 +447,4 @@
     "schemas": {},
     "tables": {}
   }
-}
\ No newline at end of file
+}
```
carletex, line 450: let's undo this formatting change

carletex:
This is looking good, Pablo! Did an initial pass. lmk what you think.

Thanks man!!

packages/nextjs/services/database/seed.data.example.ts
```diff
@@ -1,5 +1,13 @@
...
+  BatchStatus,
+  BatchUserStatus,
+  BuildCategory,
+  BuildType,
+  ChallengeId,
+  ReviewAction,
+  UserRole,
+} from "./config/types";
 
 export const SEED_DATA_VERSION = "1.1.2";
```
technophile-04, line 12: We need to change the number

packages/nextjs/services/database/config/schema.ts
technophile-04: Woahhh love this!!! Din't knnew about this! 

So as I understand we can have this function `convertFirebaseIdToUuid ` function BG v3.5 and also in the import builds scripts?

packages/nextjs/services/database/config/schema.ts
technophile-04: Ohhh was just researching about it and enum seems good so this case. Since we know we have limited things and it won't grow exponentially. Like only down side of this is whenever we will need to add category or type we need to create a migration. Which seems fine to me. 

Also with enum, filtering with faster as compared to having tables and joining them 

> I guess the enum returns an array?

Like you mean to say internally in DB or example the TS `BuildCategory` enum in code? 

If in DB, like postgres seems to have native enum type, which will restrict a particular column to certain strings. and it JS runtime it actually object if you console log it

packages/nextjs/services/database/config/schema.ts
technophile-04: > Not sure, especially about the "name" one. Wondering if we should use tsvector for full search (if we really want to use it)
>
> Maybe we could start only by buildtype / category and review on the future when thinking about name / description search.

Yup agree that we can remove the index from name and just have it for buildType, category. 

For name we should be able to query with `isLike` similar to how we did [here in batches](https://github.com/BuidlGuidl/SpeedRunEthereum-v2/blob/e628441dfc2b6a5c278138f80878163b9a32197b/packages/nextjs/services/database/repositories/batches.ts#L30) 

Didn't knew about `tsvector` soooo cool! Yeah like it would be nice in future to have it for name and description as drizzle [demonstrated here](https://orm.drizzle.team/docs/guides/postgresql-full-text-search) love it!

packages/nextjs/services/database/config/schema.ts
technophile-04: > Do we need this build_builder_build_idx index?  

Ohh actually this index seems redundant. Was just learning about it and you are correct! Like since here they we have created composite key (buildId, userAddress) when we do this, postgreass actually creates index of `buildId-buildeAdress` + also the index on `buildId` (left column of composite key). So we can remove it 🙌 

So the above index helps us in doing: "Fetching a Build WITH its Builders"

We need to keep the other index on `build_builder_user_idx`  when we want to do: "fetch all builds from a single builders" (since right column in composite key is not indexed by default)

packages/nextjs/services/database/config/schema.ts
technophile-04: > we could rename as stars, but maybe it gets confused with GH stars?

lol this Pablos comment makes sense, tried thinking of better name but lol like seems good only....but yeah stars also kind of make sense...was thinking of bookmarks but it's give that ranking feel when people come to builds page

packages/nextjs/services/database/config/schema.ts
carletex: > So as I understand we can have this function convertFirebaseIdToUuid  function BG v3.5 and also in the import builds scripts?

Yep. For now we'd only need it in import script. But if we want to redirect builds URLs from BG to SRE, we could have the same function

packages/nextjs/services/database/config/schema.ts
carletex: Oh, I wasn't asking about enum (we already dig on them on a previous PR)... I was asking about the change Pablo made.

from

```
export const reviewActionEnum = pgEnum("review_action_enum", [
  ReviewAction.REJECTED,
  ReviewAction.ACCEPTED,
  ReviewAction.SUBMITTED,
]);
```

to

```
export const reviewActionEnum = pgEnum("review_action_enum", ReviewAction);
```

> I guess the enum returns an array?

I said this because on the main version we are doing an array, but enums are declared as on object. Just was a bit confusing for me!

technophile-04: Ohhhhh yes missed it! Yeah like in latest version they seem allow passing the whole TS enum (which is object) instead of earlier just the array of string values 

[In this version](https://github.com/drizzle-team/drizzle-orm/releases/tag/0.42.0)

Like all the docs example used array of strings but people requested that feature and I think they updated it. 

In the code they seem to have overload for function [`pgEnum`](https://github.com/drizzle-team/drizzle-orm/blob/ad28dcd494d043fc39fa15a1622bb7a51deb6090/drizzle-orm/src/pg-core/columns/enum.ts#L144-L152) 

and internal implementation they do (`Object.values(wholeEnum)`) so its converted to array of string finally. 

Also in the new migration generated I don't see any changes to other table so it's generating same pg enum finally 🙌 

Maybe this way is more easy and nice to write

technophile-04 (commented):
TYSM Pablo!! Apart from carlos suggested changes this is looking really good to me! 


Lol for me `build_builders` table was bit non-intuitive but learning more it indeed seem nice optimal way of doing things having a "junction table" since there was many-many relation b/w users and builds and also we wanted `isOwner` which nicely fits into this relation. 

Everything looks nice to me!!

packages/nextjs/services/database/config/schema.ts
Pabl0cks: Love this! It's brilliant 🔥

packages/nextjs/services/database/config/schema.ts
Pabl0cks: Sorry! I really need to do a proper deep dive on indexes before dropping them all over the place 🙏

Pabl0cks:
Thanks for the review, research and suggestions @carletex and @technophile-04! 

Will try to have everything properly solved tomorrow 🙌

packages/nextjs/services/database/migrations/meta/0003_snapshot.json
Pabl0cks: Fixed in e9043aa143c9bf1a6fbeb783de297c83d19d67da

packages/nextjs/services/database/config/schema.ts
Pabl0cks: Not the best naming 😅 Fixed 7020a8094861e536a1191fe891577a03871ea8a3

packages/nextjs/services/database/config/schema.ts
Pabl0cks: Leaving it as `likes` 🙏

packages/nextjs/services/database/config/schema.ts
Pabl0cks: Removed the useless index in 9abab2cca6ee365f56563d6e83400fa86f813aa9 🙌

Thanks for the insights and learnings ♥

packages/nextjs/services/database/config/schema.ts
Pabl0cks: Removed the name index for now in 9abab2cca6ee365f56563d6e83400fa86f813aa9 

As you suggest, can revisit in the future with `tsvector` or other approach if we decide to aim for full search, loved to discover Postgres had this feature!

packages/nextjs/services/database/config/schema.ts
Pabl0cks: > So we could use a deterministic function that converts firebaseID to uuids (on the import script). Something like:

Tested it and is working nicely, TYSM!

packages/nextjs/services/database/seed.data.example.ts
Pabl0cks: Updated in e2f7a553df4209459813cd4c0bf4dc76539eebb4 🙌

Pabl0cks:
Just pushed the changes to address the feedback, let me know if I missed anything or any other tweak to tackle, thanks again!! 🙌

> Best way to store internally if a build uses Scaffold-ETH and if it's an extension, we use this data for our scaffoldeth.io extension list, and could have a filter in the builds page too for both fields.
>   - Maybe having 2 extra fields (boolean) for Scaffold-ETH and Extension? In the UI we could show them unchecked if you select "Dapp" as the build_type?

Should we tackle this last item in a future PR?

packages/nextjs/services/database/config/schema.ts
carletex: thanks Shiv!

carletex (approved):
Looks great Pablo! Thanks.

I'll merge and run the migration in production.

> Should we tackle this last item in a future PR?

Yes, not sure what we want to do with this yet, so we can discuss in another PR/issues.

I think next: are #140 and #139. We can tackle them in parallel. I might tag Rinat for 139

carletex:
> Lol for me build_builders table was bit non-intuitive but learning more it indeed seem nice optimal way of doing things having a "junction table" since there was many-many relation b/w users and builds and also we wanted isOwner which nicely fits into this relation.

Yes, this is tricky! But it's the way to go when you have `N:N` relations.
- `1:1` (just include the data in a single  table. e.g.  builder + socialGithub => builder table)
- `1:N` (create 2 tables referencing each other from N to 1: e.g. builder + challenge => builder + userChallenges tables)
- `N:N` (create 3 tables: the extra table holds the multiple references to the other two + you can add "metadata". e.g builder + build => builder + build + buildBuilder tables)

PR #137, opened 2025-04-18
Batches page

rin-st:
Batches Page

- batches infinite table
- possibility to search by name
- creation of the batch
- editing the batch

Also, 
- moved `InfiniteTable` to separate component and used it for both batches and users pages
- used one component for both modals
- only admin can update/create batches
- if non-admin tries to see `/batches` page for now `null` returned. We need to figure out how to restrict only-admin client pages, I think for different PR
- if some required field is not filled showing notification
- updated tg field in schema to `notNull` + added migration

Creating batch:

<details><summary>Details</summary>
<p>

https://github.com/user-attachments/assets/282b3cf9-249a-439b-8ef1-30d8ff79d27d

</p>
</details> 

Updating batch + some sorting:

<details><summary>Details</summary>
<p>

https://github.com/user-attachments/assets/cd0319cd-44ce-4240-ba09-8324a2eed330

</p>
</details> 

To test: 
`yarn drizzle-kit studio` and then change your role to Admin.
Check `/batches` page

First PR of #125

PR #143, opened 2025-04-22
assign batch id's to users dynamically (seed)

technophile-04:
### Description: 

This sovles the issue we were facing where we would get (duplicate primary key error)

The issue was happening because, in seed data we had hardcoded id's for batch. So PG internal autoincrment wasn't increasing when we created the batches. i.e (autoincrement is left behind in fresh case it would be 0) but still batch.Id will be 3 

But then when we try to create a batch normally without providing id, we get duplicate primary key error because it the autocrement try to create batch it id 1 but since it's created via seed script. 

This PR assign the created batch id's to initial users in order, i.e assign batch id's dynamically. So we all the states are nicely aligned and we let DB handle the the id's 

Solves https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/136#pullrequestreview-2783764910 and also https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/137#discussion_r2052555137

You will be needing to copy the seed.data.example.ts file again to seed.data.ts

### DIFF
--- a/packages/nextjs/services/database/seed.data.example.ts
+++ b/packages/nextjs/services/database/seed.data.example.ts
@@ -38,6 +38,7 @@ export const seedUsers: (typeof users.$inferInsert)[] = [
     socialEmail: "afo@wefa.app",
     socialX: "Time_Is_Oba",
     socialGithub: "Oba-One",
+    batchStatus: BatchUserStatus.CANDIDATE,
   },
   {
     userAddress: "0x45334F41aAA464528CD5bc0F582acadC49Eb0Cd1",
--- a/packages/nextjs/services/database/seed.data.example.ts
+++ b/packages/nextjs/services/database/seed.data.example.ts
@@ -1,7 +1,7 @@
 import { batches, challenges, userChallenges, users } from "./config/schema";
 import { BatchStatus, BatchUserStatus, ChallengeId, ReviewAction, UserRole } from "./config/types";
 
-export const SEED_DATA_VERSION = "1.1.0";
+export const SEED_DATA_VERSION = "1.1.1";
 
 // Using Drizzle's inferred insert types to ensure seed data
 // matches database schema requirements
--- a/packages/nextjs/services/database/seed.data.example.ts
+++ b/packages/nextjs/services/database/seed.data.example.ts
@@ -202,7 +202,7 @@ export const seedChallenges: (typeof challenges.$inferInsert)[] = [
 
 export const seedUserChallenges: (typeof userChallenges.$inferInsert)[] = [
   {
-    userAddress: "0xB4F53bd85c00EF22946d24Ae26BC38Ac64F5E7B1",
+    userAddress: seedUsers?.[0]?.userAddress,
     challengeId: ChallengeId.SIMPLE_NFT_EXAMPLE,
     frontendUrl: "https://dreary-use.surge.sh/",
     contractUrl: "https://sepolia.etherscan.io/address/0x7f918d7b7d0fe0d3a8de3c0570ed4e154c0096e0",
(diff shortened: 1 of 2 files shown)


packages/nextjs/services/database/seed.ts
```diff
@@ -83,10 +83,19 @@ async function seed() {
...
     // Insert fresh data
     console.log("Inserting batches...");
-    await db.insert(batches).values(seedBatches).execute();
+    // Insert batches and get their generated IDs
+    const insertedBatches = await db.insert(batches).values(seedBatches).returning();
+
+    // Assign each batch to a starting user in order
+    const updatedSeedUsers = seedUsers.map((user, idx) => {
+      if (idx < insertedBatches.length) {
+        return { ...user, batchId: insertedBatches[idx].id };
```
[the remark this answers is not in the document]
carletex: An alternative to this solution:

Add an extra field "builders" to `seedBatches` (we'd just need to tweak the type)

e.g.

```tsx
// Fix type
export const seedBatches: (typeof batches.$inferInsert)[] = [
  {
    name: "Batch 1",
    startDate: new Date(1705315180000),
    status: BatchStatus.CLOSED,
    telegramLink: "https://t.me/joinchat/1",
    contractAddress: "0x0000000000000000000000000000000000000000",
    users: ["0x0...0", "0x1....0"]
  },
```

and then on this seed script, we just update the field for those users.

rin-st (approved):
Lgtm, thanks!

packages/nextjs/services/database/seed.data.example.ts
```diff
@@ -364,25 +361,25 @@ export const seedUserChallenges: (typeof userChallenges.$inferInsert)[] = [
...
+    userAddresses: [seedUsers?.[0]?.userAddress, seedUsers?.[1]?.userAddress],
   },
   {
-    id: 2,
     name: "Batch 2",
     startDate: new Date(1707820780000),
     status: BatchStatus.CLOSED,
     telegramLink: "https://t.me/joinchat/2",
     contractAddress: "0x0000000000000000000000000000000000000000",
+    userAddresses: [seedUsers?.[2]?.userAddress, seedUsers?.[3]?.userAddress],
```
carletex, line 380: Can you add `batchStatus` to `seedUser[3]`? It's `null` now.

I guess this is something we could fix on the database too (batch status should probably be notNull and default to candidate)

carletex (commented):
Love this @technophile-04! We can use this pattern for future stuff where there are relantionships (like builds)

Love the use of `seedUsers?.[0]?.userAddress,` (I was searching and replacing on my seed.data to use my address): Could you update the `seedUserChallenges` rows to use it too? (instead of hardcoded addresses)

packages/nextjs/services/database/seed.data.example.ts
technophile-04: Ohh nice catch! Thanks updated it at https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/143/commits/0308c1fdcd03ebbb706af1499e2b57ba96e75284

technophile-04:
> Could you update the seedUserChallenges rows to use it too?

Ohh thanks! Yes updated at https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/143/commits/9002f9c7176c3648bd677d5c00ccc67be7d7aa0a 🙌

packages/nextjs/services/database/seed.data.example.ts
carletex: > Like somehow on db level we need to make sure if batch_id exist then batch_status shouldn't be null at all.

Not sure if it's possible to enforce this on the database level, but yes, that would be what we need. No big deal tho!

carletex (approved):

PR #146, opened 2025-04-23
Loading spinner on builders count

carletex:
Tiny thing: show spinner instead of `0` while loading the builders on the /builders page

image

image


technophile-04 (approved):
Thanks looks great!! and works nicely!

PR #148, opened 2025-04-23
create import location and social scripts

### LINKED ISSUE #145: Import BG3.5 users
Make sure that all the Builders in BG3.5 are in SRE too.

Maybe @technophile-04  was doing it as part of #97, where we also update the existing users with updated socials / location. If that's the case (but adding it as a sub-issue for now)

technophile-04:
### Description: 

1. Check if the user is present or not. If not creates it and updates the social + localtion
2. If the user is present, then we update the location and soicals always

fixes #145 

And 2nd part of #97


packages/nextjs/services/database/bg-imports/20250423_location_socials.ts
```diff
@@ -0,0 +1,101 @@
...
+    cap?: string;
+    lastIndexedBlock?: number;
+    balance?: string;
+    streamAddress?: string;
+    frequency?: number;
+  };
+  [key: string]: any;
+}[];
+
+export async function importLocationAndSocials() {
```
rin-st, line 43: nit: why to export it?

packages/nextjs/services/database/bg-imports/20250423_location_socials.ts
```diff
@@ -0,0 +1,101 @@
...
+        });
+
+        // Prepare socials mapping
+        const socials = {
+          socialTelegram: builder.socialLinks?.telegram || undefined,
+          socialX: builder.socialLinks?.twitter || undefined,
+          socialGithub: builder.socialLinks?.github || undefined,
+          socialInstagram: builder.socialLinks?.instagram || undefined,
+          socialDiscord: builder.socialLinks?.discord || undefined,
+          socialEmail: builder.socialLinks?.email || undefined,
```
rin-st, line 66: nit: all that `|| undefined` not needed

rin-st (approved):
Working great, thanks!

Added two small nitpicks

packages/nextjs/services/database/bg-imports/20250423_location_socials.ts
technophile-04: Ohh we had it in batch data import script too : 

https://github.com/BuidlGuidl/SpeedRunEthereum-v2/blob/f87a9ac814b56bfc3b99680d170ef21c76ab4292/packages/nextjs/services/database/bg-imports/20250416_batch_data.ts#L32
 
 But happy to remove it if we feel right that way 🙌

packages/nextjs/services/database/bg-imports/20250423_location_socials.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export async function importLocationAndSocials() {
+  try {
+    console.log("==== Importing Location and Socials ====");
+    console.log("Fetching data from BG firebase...");
+    const response = await fetch(BG_USERS_ENDPOINT);
+    const data: BGBuildersApiResponse = await response.json();
+    console.log(`Fetched ${data.length} users`);
+
+    await db.transaction(async tx => {
```
carletex, line 51: Thinking if it makes sense to do a transaction here? I think this can get slow / painful rollbacks. 

We are basically going user by user.... so if one of them fails, it doesn't make sense to rollback all of them, right?

I think what we really need is "batches", (which is supported by Neon, But not for Local postgress), where instead of sending a bunch of request, it's only one (or a few, if we make chunks)

PS: Sorry for being a pain with this database stuff, but I prefer if we research and learn a lot now, so we can use that for the future

packages/nextjs/services/database/bg-imports/20250423_location_socials.ts
```diff
@@ -0,0 +1,101 @@
...
+import { lower, users as schemaUsers } from "../config/schema";
+import * as dotenv from "dotenv";
+import { eq } from "drizzle-orm";
+import * as path from "path";
+
+dotenv.config({ path: path.resolve(__dirname, "../../../.env.development") });
+
+const BG_USERS_ENDPOINT = `${process.env.NEXT_PUBLIC_BG_BACKEND}/builders`;
+
+export type BGBuildersApiResponse = {
```
carletex, line 11: Would it make sense to have a bg types file somewhere? (maybe inside bg-imports? and move the batch data stuff too) 

Since we will be using it in different places.

packages/nextjs/services/database/bg-imports/20250423_location_socials.ts
technophile-04: fixed at https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/148/commits/f87a9ac814b56bfc3b99680d170ef21c76ab4292

packages/nextjs/services/database/bg-imports/20250423_location_socials.ts
technophile-04: Ohh no no, yes completely agreee that this will help us a lot in future! Will research about it 🙌

packages/nextjs/services/database/bg-imports/20250423_location_socials.ts
technophile-04: Ohh yes moved it inside `types.ts` file in `bg-imports` at https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/148/commits/d4adecc9ef25894468eca222cfa5d6d74dc17969

packages/nextjs/services/database/bg-imports/20250423_location_socials.ts
technophile-04: removed at https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/148/commits/526711a4dc0cfc8e39750967bfab6d38912a4ceb

packages/nextjs/services/database/bg-imports/20250423_location_socials.ts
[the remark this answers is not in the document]
technophile-04: Updated it to be [users_profileCard_data](https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/148/commits/12b1ee524948645c28cf6770b4583448dd924ccc)

packages/nextjs/services/database/bg-imports/20250423_location_socials.ts
carletex: One thing it would be nice to try (for the sake of learning):

=> Prepare the loop with the queries

If NEON: 
- wrap all of them in db.batch
- Try to make chunks of 500 / 1k users

If non-NEON:
- just run them as simple queries (no need for TX either)

---

Also, if we like this.... it might make sense to try to set up neon locally (even if it's a bit ugly haha)

packages/nextjs/services/database/bg-imports/20250423_location_socials.ts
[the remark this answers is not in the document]
carletex: Ok, proposing something:

What if we change the naming to `0001_`, so imports will be executed in the right order and don't have to think it too much (if we run them in 2 weeks :D)

IMO:

- 0001_user_data.ts (the import script from this PR)
- 0002_batch_data.ts (from #136 )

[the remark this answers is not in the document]
technophile-04: Noiceee yess love this idea!! Renamed them at https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/148/commits/e3aa56b14812a95ab8bc402b0d5664f60943f0fb 🙌

packages/nextjs/services/database/bg-imports/20250423_location_socials.ts
technophile-04: Just updated the script (0001_user_data.ts) to have `db.batch` when it's available. 

> Try to make chunks of 500 / 1k users

Didn't do it since the BG builders are only 1340 (appx 1k) and 

[image](https://arc.net/l/quote/mnuzpyob)

So it should be fine? 

Also shall we update the `0002_batch_data.ts` too? I think it should be fine because data won't be too much but maybe for consistency? Happy to update it as well 🙌

packages/nextjs/services/database/config/postgresClient.ts
```diff
@@ -63,6 +74,10 @@ const dbProxy = new Proxy(
       const db = getDb();
       return db[prop];
     },
+    has: (target, prop: keyof DbProxy) => {
```
technophile-04, line 77: We had to add this so that we can do: 

`if("batch" in db)` check, otherwise it JS was failing because there was no method on proxy to do it, and it was resulting in false always.

technophile-04:
While discussion this with @carletex  we thought about https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/148/commits/ccadf45ae712778c212e64468af43bb8a83a9922  and added a `executeQueries` (we can change the name) function to db proxy. Which is a smart function and executes the queries passed to it on as separate queries or in batch. 

This is helpful in scripts, well it's was not that helpful in batches script (check #155) but it will be helpful for future scripts?

rin-st (approved):

packages/nextjs/services/database/bg-imports/20250423_location_socials.ts
rin-st: > Thinking if it makes sense to do a transaction here? I think this can get slow / painful rollbacks.

> We are basically going user by user.... so if one of them fails, it doesn't make sense to rollback all of them, right?

I'm not working often with dbs so my understanding/some silly questions

Why rollbacks are painful? Aren't they work automatically? Well, I've asked AI and read about locking tables and hence other queries, additional i/o operation to revert and some more reasons . Just want to understand what is the main issue in our case

If I'm understanding right, its possible in batches to fail only part of the queries. So if we use batches, we need a mechanism that will help us 
1. Notifications system of failing queries. So we understand it quickly, not in some months. Hope it works in neon by default. 
2. Save logs somewhere to understand which queries failed. As I see neon has Monitoring/query history but I'm not sure it's convenient to analyze and work with that logs if lots of queries failed
3. Sometimes just retrying queries could help so probably we need that mechanism too

Or plan just to: update db -> check neon for errors -> try to find the error -> revert/retry ?

> I think what we really need is "batches", (which is supported by Neon, But not for Local postgress), where instead of sending a bunch of request, it's only one (or a few, if we make chunks)

As I understand main reason to use batches is when performance matters. But in our case it's just one-time operations, do we need to care about it? Or maybe using transactions has some cost?

carletex: We are being a bit nitpicky with the DB stuff (but because it's fun, and we want to learn more haha), so don't worry too much. But let me try to answer some questions

> Why rollbacks are painful? Aren't they work automatically? Well, I've asked AI and read about locking tables and hence other queries, additional i/o operation to revert and some more reasons . Just want to understand what is the main issue in our case

Imagine you want to import 100k rows (user data) into the database. If you use a `transaction` without chunks, if the row 99.999 fails (e.g. invalid data), the "automatic" rollback will happen... and if we are doing this in a production database, it can kill it / make it super slow (people using the platform will be affected)

That's why I asked: does it make sense to use tx here? We don't really need to rollback in this case (if a user fails to import, we can check what was the error... and reimport/fix that user only... no need to revert back the others). This is still hardcore (100k requests!)

That's why, for this particular use case... `batch` is the way to go (a single TX with all the data). You can also split into chunks, but since we only have ~1k users, no need.

Note: Transactions make sense when you want "all or nothing".... e.g. if one query fails it will leave the DB in a incoherent state

> Notifications system of failing queries

I haven't thought about this, but it's important. What happens if 1 query fail? Do we get a clear message in the response?

cc @technophile-04 

> As I understand main reason to use batches is when performance matters. But in our case it's just one-time operations, do we need to care about it? Or maybe using transactions has some cost?

And yes, you are right about this. We are just geeking around :nerd_face:

rin-st: Thanks for the answer!

> Imagine you want to import 100k rows (user data) into the database. If you use a transaction without chunks, if the row 99.999 fails (e.g. invalid data), the "automatic" rollback will happen... and if we are doing this in a production database, it can kill it / make it super slow (people using the platform will be affected)

Thought about similar case but with batches. Imagine you want to import 100k row using batches, and random 1k imports fails. And fixing this also could be a pain. But yeah, probably it's much easier to just revert the changes chunk by chunk

For me both solutions looks good for now since we have not so much rows, maybe batches a bit better

technophile-04: > does it make sense to use tx here? We don't really need to rollback in this case (if a user fails to import, we can check what was the error... and reimport/fix that user only... no need to revert back the others). This is still hardcore (100k requests!)

Ohh like if the batch fails then also the whole thing rollsback. 

The only advantage of `batch` over `tx` is we are saving number of roundtrip b/w the script and database. 

So for example(using transaction): 

```ts
  await tx.update(...);
  await tx.update(...);
});
``` 

Is sent like this: 

```sql
BEGIN; <-- one round trip
UPDATE table1 SET column_name1 = 'new_value' WHERE conditions; <-- 2 round trip
UPDATE table2 SET column_name2 = 'new_value' WHERE condition2; <-- 3rd round trip
.... <--- n+1 round trip
COMMIT;(or rollback happens) <-- final round round trip
```

But when we do batch: 

```ts
 db.batch([
   db.update(...),
   db.update(...)
 ]); 
```

It deposits whole thing to database in one round trip: 

```sql
|-------------------------------------------------------------------
| BEGIN;
| UPDATE table1 SET column_name1 = 'new_value' WHERE |conditions;
|UPDATE table2 SET column_name2 = 'new_value' WHERE |condition2;
|.... <--- n+1 query
|COMMIT(or rollback happens);  final query
-------------------------------------------------------------------
```

So this whole box is handed as one SQL statement itself.  (source [here](https://github.com/drizzle-team/drizzle-orm/issues/3828))

So like `db.batch()` [internally uses transaction only](https://neon.tech/docs/serverless/serverless-driver#issue-multiple-queries-with-the-transaction-function) and [2](https://orm.drizzle.team/docs/latest-releases/drizzle-orm-v0294#new-features) its just that we are saving round trips. 

lol sorry for overexplaining the stuff and maybe you guys already got it 

> I haven't thought about this, but it's important. What happens if 1 query fail? Do we get a clear message in the response?

So like if any of the query fails in batch it ends the script and log this: 

<details>

<summary>
    Failed query log
</summary>

Screenshot 2025-04-29 at 12 07 00 PM


</details>

and rollbacks the whole thing, so there are no changes in neon db.

carletex: Thanks for looking this up, Shiv!

Then `batches` is definitely the way to go (when possible), right?

technophile-04: > Then batches is definitely the way to go (when possible), right?

Ohh yes 🙌

packages/nextjs/services/database/bg-imports/0001_user_data.ts
```diff
@@ -0,0 +1,78 @@
...
+              ...socials,
+              location: builder.location || null,
+              updatedAt: new Date(),
+            })
+            .where(eq(lower(schemaUsers.userAddress), userAddress.toLowerCase()));
+        } else {
+          return db.insert(schemaUsers).values({
+            userAddress,
+            ...socials,
+            location: builder.location || null,
```
carletex, line 61: For new users, should we keep `createdAt` from BG3.5 or not?

packages/nextjs/services/database/bg-imports/0001_user_data.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+  } catch (error) {
+    console.error("Error during data import:", error);
+  } finally {
+    await db.close();
+    process.exit(0);
+  }
+}
+
+importLocationAndSocials();
```
carletex, line 78: nit: I think this could just be `importData` to unify all the scripts. Less work when copy / pasting this :D (and the file name / description should be enough to identify the script)

carletex (approved):
This is working great, Shiv!

Added a few nits that we could tackle

packages/nextjs/services/database/bg-imports/0002_batch_data.ts
```diff
@@ -0,0 +1,102 @@
...
+
+const BG_BATCHES_ENDPOINT = `${process.env.NEXT_PUBLIC_BG_BACKEND}/batches`;
+const BG_USERS_WITH_BATCHES_ENDPOINT = `${process.env.NEXT_PUBLIC_BG_BACKEND}/builders/batches`;
+
+async function importData() {
+  try {
+    console.log("==== Importing BATCH DATA ====");
+    console.log("Fetching data from BG firebase...");
+
+    const batchesResponse = await fetch(BG_BATCHES_ENDPOINT);
```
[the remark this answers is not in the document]
rin-st: in https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/154 I sorted them by `startDate` since as I understand names could be different in future (universities etc), so probably it will be easier to sort by that field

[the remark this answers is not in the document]
carletex: 100%!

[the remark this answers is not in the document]
technophile-04: Noicee yup agree, fixed at https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/148/commits/ad57c96adaf6832a3abdd4fd8d7319cb52a02ce0 

image


lol my local DB autincrement didn't reset after I removed the data hence it started from 21 but it's sequential 🙌

packages/nextjs/services/database/bg-imports/0001_user_data.ts
technophile-04: lol yup fixed at https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/148/commits/bcb07ba6dc20dd2b98eece465acd342ae4a5c912 🙌

packages/nextjs/services/database/bg-imports/0001_user_data.ts
technophile-04: Ohh noicee yes agree fixed at https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/148/commits/a0402c062b085fffc15b2f69070072b498fada8e 🙌

carletex (approved):
Amazing stuff! Tested it locally and by importing to staging .. and it works perfect.

Merging! (won't run them in production yet)

PR #152, opened 2025-04-25
fix: table header

rin-st:
After last table update, header background was removed

This micro pr fixed this

before:
Screenshot 2025-04-25 at 12 18 33

after:
Screenshot 2025-04-25 at 12 18 51


carletex (approved):
Thanks Rinat!

PR #153, opened 2025-04-28
SEO initial improvements (challenges and homepage metadata, robots.txt and sitemap.xml)

Pabl0cks:
Creating this initial PR with a few "low-hanging fruit" SEO improvements:

- Add SEO metadata for challenge pages and the homepage.
- Add a static `robots.txt` to exclude `/builders/` (and its children) from crawling. Currently, we don't have any visible links to `/builders` for crawlers, but if we ever do, we'd wake up with thousands of URLs crawled for SRE builders, which wouldn't have a good effect on our SEO (solidity development oriented).
- Add a static `sitemap.xml` with challenge slugs to try to index all our challenges, even the disabled ones for new builders (and for SEO crawlers).

A few notes:

- The layout approach felt like the right way to go for challenges metadata, but let me know if you think I should follow a different approach. 🙏
- I went for a simple approach for `robots.txt` and `sitemap.xml`, since the size of the website is manageable. However, we could reconsider moving to a dynamic approach if we ever feel the need.
- Regarding the last change, to make all (active) challenges accessible to crawlers, I can create another issue to discuss whether we should enable all challenges on the homepage and show feedback on the challenge pages to let users know which ones they're missing in order to submit the current challenge.  
We could also consider removing all requirements and simply adding a visual suggested path when we redesign the builders page.


carletex (changes requested):
As we are talking live on a call:

- move challenge metadata to a file (we could use it with #169 related guides too)
- use our own generate metadata
- "ethereum" on layout meta description

Pabl0cks:
Thx for the feedback @carletex !! 

Updated where we store and fetch the challenges metadata in https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/153/commits/eb383d9a956b1e430926a263150075fa070b84f3 and tweaked the generic meta description adding back "ethereum" in https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/153/commits/be45becade90f472ff741237e06e1bb11e8d563e 🙌

carletex (approved):
Good stuff, Pablo, thank you!!!

As you said, we can consider moving some of these to the database (maybe when we have more challenges)

Issue #45, opened 2025-03-12
Cache ENS


carletex:
Let's cache ENS, saving it to the database.

1. Let's save it when the user register (if they have)

and then we could:

a. Cron job to check empty ENS every 6 (or whatever) hours
b. Check for ENS when the user with empty ENS submit something / makes an action (like a middleware that runs when the user doesn't have an ENS)
c. Have a "refresh" button in the user profile, so they can click it and we'll fetch it and cache it.

`c` might be good (it's the most efficient at least haha), especially if we have the ENS stuff as a side quest (get your ENS! Read this article on what ENS is and how to get One... etc)

rin-st:
what do you think about this way:
when user is logged in we use `useUser` hook. If there's no ens we use [useEnsName](https://wagmi.sh/react/api/hooks/useEnsName) and if ens exists we add it automatically.

So we don't need all the steps from initial issue message 

> c might be good (it's the most efficient at least haha), especially if we have the ENS stuff as a side quest (get your ENS! Read this article on what ENS is and how to get One... etc)

Regarding this point if user already logged in and just added ens we still can add Refresh button but it makes sense only until next login, otherwise it will set automatically

PR #154, opened 2025-04-28
Builders in batches page

rin-st:
<details><summary>How it works</summary>
<p>

https://github.com/user-attachments/assets/01503bbf-32bd-434c-83f3-a485e120fe4e

</p>
</details> 

notes:
- "status" and "builds" columns were not added for now, need to add that fields after imported
- also, `user.creationTimestamp` should be imported for correct sort of last column. ✅  this is fixed
- when sorting by batch column, batches sorted by `startDate`. Because as I understand names could be different than just "Batch N" in future
- still the same "if not admin return null" check as on client page
- a bit updated `InfiniteTable`s `queryFn` prop

To test:
- change your role from drizzle studio to admin (register and then sort by createdAt in studio if imported all the users)
- check `/admin/batch-builders`

Do I need to wait for all the imports? Or could it be done in #148 , except builds (`user.status` and `user.creationTimestamp`)?


phipsae:
@rin-st I will have a look at it tomorrow. Sorry for delay.

packages/nextjs/services/database/repositories/users.ts
```diff
@@ -84,6 +85,68 @@ export async function getSortedUsersWithChallengesInfo(start: number, size: numb
...
+  const query = db.query.users.findMany({
+    limit: size,
+    offset: start,
+    with: {
+      batch: true,
+    },
+    where: users =>
+      and(
+        isNotNull(users.batchId),
+        filter ? ilike(users.userAddress, `%${filter}%`) : undefined,
```
carletex, line 111: Future: We would be able to improve this if we do #45 , and we could also search in socials.

carletex (approved):
(Went ahead and reviewed this since it feels it was staling on reviews.)

This is working great Rinat, thanks!

I pushed some tweaks:
- change green badge meaning: instead of OPEN/current batch (as in /admin/batches) it shows if the builder is graduated or not.
- some spacing on tables
- Removed ToDo

There is one missing piece that is important and it's the "edit builder" for admins (so they can change the status of the builder in the batch etc). Will create an issue for it.

Merging this, let's iterate/tweak on future PRs

PR #155, opened 2025-04-28
update batch query to use batch

technophile-04:
### Description: 

Use `db.executeQueries` when possible in batch script. 

Umm I think here we don't need it and transaction here  make more sense. 

The reason why transaction here make more sense is because transaction allows to run next queries based on previous conditions. 

But with `batch` it's not possible since we need to prepare the queries ahead and just pass them in the array.

### DIFF
--- a/packages/nextjs/services/database/bg-imports/0002_batch_data.ts
+++ b/packages/nextjs/services/database/bg-imports/0002_batch_data.ts
@@ -24,74 +24,70 @@ async function importData() {
 
     console.log(`Found ${batchesData.length} batches and ${usersData.length} users belonging to a batch`);
 
-    // Start a transaction
-    await db.transaction(async tx => {
-      // Save the mapping between Firebase ID and our database ID
-      const batchIdMap = new Map<string, number>();
+    // Save the mapping between Firebase ID and our database ID
+    const batchIdMap = new Map<string, number>();
 
-      // Preload existing batches
-      const existingBatches = await tx.select().from(batches);
-      const existingBatchesByTelegramLink: Record<string, (typeof existingBatches)[0]> = {};
-      existingBatches.forEach(batch => {
-        if (batch.telegramLink) {
-          existingBatchesByTelegramLink[batch.telegramLink] = batch;
-        }
-      });
-
-      // Import batches
-      for (const batch of batchesData) {
-        const newBatch = {
-          name: `Batch ${batch.name}`,
-          startDate: new Date(batch.startDate),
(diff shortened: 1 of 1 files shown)


packages/nextjs/services/database/bg-imports/0002_batch_data.ts
```diff
@@ -24,74 +24,70 @@ async function importData() {
...
+      } else {
+        const result = await db.insert(batches).values(newBatch).returning();
+        createdBatch = result[0];
+        console.log(`Imported batch: ${createdBatch.name} (ID: ${createdBatch.id})`);
       }
-    });
+
+      // Store the mapping between Firebase ID and our database ID
+      batchIdMap.set(batch.name, createdBatch.id);
+    }
```
technophile-04, line 65: Soo here the reason why we can't use batch is because, we wanted the result of the inserted batch (batch id) 


We cab kind of achieve this with `batch` since batch also returns data but its very hard to get the types right. It's very dynamic and also the way we are constructing the queries are very dynamic too. 

So I kind of falling to using the transaction for batches script and also kind of makes sense to use but let me know 🙌

carletex: Thanks for exploring this, Shiv!

> So I kind of falling to using the transaction for batches script and also kind of makes sense to use but let me know 🙌

I agree with this!

I think transactions make sense here (+ getting the rolling back out of the box). Also, this is little data, so no need for batching.

So feel free to close the PR.

packages/nextjs/services/database/bg-imports/0002_batch_data.ts
```diff
@@ -24,74 +24,70 @@ async function importData() {
...
+          batchId: batchId,
+          batchStatus: user.batch.status.toLowerCase() as BatchUserStatus,
+        })
+        .where(eq(users.userAddress, user.id));
+      userUpdateQueries.push(query);
+    }
+
+    if (userUpdateQueries.length > 0) {
+      await db.executeQueries(userUpdateQueries as any);
+    }
```
carletex, line 90: Question: If one of the queries fails.... does it reverts or not?

If not... is it possible to wrap this into db.transaction?

Just curious!

technophile-04: Ohh my bad, didn't see it just added similar comment here : 

https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/148#discussion_r2065683978

carletex (approved):

PR #158, opened 2025-05-02
Remove public RPC

carletex:
It seems that RPC requests run in parallel and it's slowly down things like ENS resolutions

### DIFF
--- a/packages/nextjs/services/web3/wagmiConfig.tsx
+++ b/packages/nextjs/services/web3/wagmiConfig.tsx
@@ -22,9 +22,7 @@ export const wagmiConfig = createConfig({
     const alchemyHttpUrl = getAlchemyHttpUrl(chain.id);
     if (alchemyHttpUrl) {
       const isMainnet = chain.id === 1;
-      rpcFallbacks = isMainnet
-        ? [http(BG_MAINNET_RPC_URL), http(alchemyHttpUrl), http()]
-        : [http(alchemyHttpUrl), http()];
+      rpcFallbacks = isMainnet ? [http(BG_MAINNET_RPC_URL), http(alchemyHttpUrl)] : [http(alchemyHttpUrl)];
     }
 
     return createClient({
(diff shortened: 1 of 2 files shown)


technophile-04 (approved):
Thanks @carletex!

technophile-04:
> It seems that RPC requests run in parallel and it's slowly down things like ENS resolutions

Tried debugging this and didn't came to any conclusion why this is happening :( no viem discussion on this too. 

But with viem and wagmi (internally using tanstack queyr) like they are lot of variables which are playing together and might be related to it. 

--- 

I was also why it was taking soo much on cloudfare rpc, but not when I use the "https://eth.llamarpc.com" public RPC. 

For better debugging I turned off `retry: false` in both ENS hook of `Address` components, so that we don't send request when read reverts. 

It seems that cloudlare rpc when read reverts it gives "Internal server error": 

image
 
 Where as other RPC's even public rpc eth.llamarpc.com gives: 
 
image


There are retry at two levels one at [transport level](https://viem.sh/docs/clients/transports/http#retrycount-optional) and the other at [hook level](https://wagmi.sh/react/api/hooks/useEnsName#retry). We turned of hook level transport (which was making retries if the reads revert) 

But now the HTTP transport from veim seem to make request when request fails (non 200 status code) 

Screenshot 2025-05-02 at 11 04 14 PM

The above SS I took with : 

1. Having only one transport HTTP("https://eth.llamarpc.com") in `wagmiConfig` no fallback just single `transport: http("https://eth.llamarpc.com")`
2. the retry flags set to false in Address component for all the hooks. 

But when I switch to cloudflare RPC (default) in `wagmiConfig.ts` then I get: 

Screenshot 2025-05-02 at 11 08 26 PM

The above SS I took with : 

1. Having only one transport HTTP() in `wagmiConfig` no fallback just single `transport: http()`
2. the retry flags set to false in Address component for all the hooks. 

So for internal server contract call fails viem seems to retry with lots of request. 

And since we were making lot's of requests (with cloudfare) maybe that's why were getting 15s delay

PR #159, opened 2025-05-05
admin auth

technophile-04:
### Description: 

We now gate all the `/admin/*` route by having authorization check at `/admin/layout.tsx` so that only user with `"admin"` is allowed to view those pages. 

In API routes only added authorization check at `api/batches/sorted` (GET) request since other `api/batches/create | update` we are getting them with EIP712 signature. 

To test get request: 

```shell
curl -X GET "http://localhost:3000/api/batches/sorted?start=0&size=10&sorting=%5B%7B%22id%22%3A%22createdAt%22%2C%22desc%22%3Atrue%7D%5D&filter="
```

### UI Testing: 

1. On home page (http://localhost:3000) you can register yourself. 
2. go to https://local.drizzle.studio/ and update that user with "ADMIN" role. 

There are only two admin routes http://localhost:3000/admin/batches and http://localhost:3000/admin/batch-builders 

you can go to any of those route and you will "Sign In with Etherum" button. Which will prompt you to sign a siwe message and set session in the browser. 

If you disconnect wallet the session is gone. 

### Technical details: 

Using the [`viem/siwe`](https://viem.sh/docs/siwe/actions/verifySiweMessage) instead of [siwe](https://docs.login.xyz/integrations/nextauth.js) library (which seems to be not updated and requires `ethers` dependency) 

found the above while researching [here](https://github.com/rainbow-me/rainbowkit/pull/2204) 

We set the session only in `/admin/*` route. Hence the user is not prompted to sign siwe message in any other routes while connecting the wallet. 

### Next steps: 

To make it work on Vercel we need to add this env variable in vercel: 

```
NEXTAUTH_URL=http://localhost:3000 # shall be changed to latest domain
NEXTAUTH_SECRET=somereallysecretsecret
```

We can use 

```
openssl rand -base64 32
``` 

or : `npx auth secret` cli to generate that secret 🙌


packages/nextjs/utils/auth.ts
```diff
@@ -0,0 +1,127 @@
...
+          return null;
+        }
+
+        const user = await getUserByAddress(siweMessage.address);
+
+        if (!user) {
+          return null;
+        }
+
+        return { id: user.userAddress, role: user.role };
```
technophile-04, line 84: NOTE: Here we don't create user unlike other apps. We just set the session if the user is present and if not we don't set the session at all. 

Since this is for just admin, we preassume they are already present in DB

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton/AddressInfoDropdown.tsx
```diff
@@ -116,7 +117,10 @@ export const AddressInfoDropdown = ({
             <button
               className="menu-item text-error btn-sm !rounded-xl flex gap-3 py-3"
               type="button"
-              onClick={() => disconnect()}
+              onClick={() => {
+                disconnect();
+                signOut();
```
rin-st, line 122: Same problem as https://github.com/BuidlGuidl/ens-pg/pull/55 . Additional info in tg

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton/index.tsx
```diff
@@ -21,6 +24,14 @@ export const RainbowKitCustomConnectButton = () => {
   const { address: connectedAddress } = useAccount();
   const { data: user, isLoading: isLoadingUser } = useUser(connectedAddress);
   const { disconnect } = useDisconnect();
+  const router = useRouter();
+  const { isAuthenticated } = useAuthSession();
+
+  useEffect(() => {
```
rin-st, line 30: This causes infinite reloadings for notFound pages, see example with `/asdf`

https://github.com/user-attachments/assets/14dc3def-81fd-4d43-afb4-e3a1ddaa0585

rin-st (approved):
> found the above while researching https://github.com/rainbow-me/rainbowkit/pull/2204

Great find!

And great job, working good! Found two issues

rin-st:
And one more question, should we restrict also `api/users/in-batches` since it's used only in admin page?

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton/index.tsx
technophile-04: https://github.com/user-attachments/assets/5b5ec66d-e7b3-43f1-9772-42b5f5b2bb4c

For me it's not happening, like I see you disconnected from wallet. Ideally if you disconnect the `isAuthenticated` will be false only right? Like even if you disconnect from wallet UI 

The logic from https://github.com/BuidlGuidl/ens-pg/pull/55 will remove the session and `isAuthenticated` will be false for sure 🙌

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton/AddressInfoDropdown.tsx
technophile-04: Fixed at https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/159/commits/70dbec23c4c835b8b618e3a83c32d24d1125ef19 

Thanks 🙌

technophile-04:
Just pushed https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/159/commits/96c4e6ac598d22c786249fc2b2789f1024b2acf9 was talking with Carlos and realized it's more inututive flow to show "Acess Denied" instead of redirecting to home page. So that admin can connect to his correct wallet. 


https://github.com/user-attachments/assets/495c71f0-bd78-4d45-bfd1-7ff2172ecbab

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton/index.tsx
rin-st: My bad it reloads when connected as admin. Connect button doesn't show it since it's always loading after page reload

rin-st: rechecked, works as expected!

technophile-04: Ohhhhh I see just fixed https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/159/commits/9e14f13abd5f80c302b7e7f18cbb8b0b0a56b417 


And since we know that admin will authenticate on `/admin/*` routes only. Since we have `router.refresh()` in `SingIn` component it will fetch the admin data. 

--- 

The problem with Not Found page is that it doesn't seem to remember the session when we do `route.refresh()`. While other pages when `isAuthenticated` becomes true and we do `router.refresh()` the `isAuthenticated` remains true (and since useEffect dependency does not change we don't get the infinite loop)

Pabl0cks (approved):
GJ Shiv! LGTM!

Before last commits I was facing some issues on my first login, it was redirecting me to the homepage without showing the SIWE login, but now is working great 🙌🔥

carletex:
> NEXTAUTH_URL=http://localhost:3000 # shall be changed to latest domain

On https://next-auth.js.org/configuration/options#nextauth_url
image

Let's see if it works after merging, or I if need to enable something.

> NEXTAUTH_SECRET=somereallysecretsecret

Generated two for this: Production & Non-production secrets

PR #162, opened 2025-05-07
Add Builds import script with basic categorization

Pabl0cks:
Adds a new import script (`0003_build_data.ts`) that pulls build data from the BuidlGuidl API and imports it into our database. It uses the `aux_builds_categorization.csv` file (generated with Gemini 2.5 Pro\*) to normalize build types and categories.

The script is probably over-commented and over-logged — can clean it up after the first round of tests 🙏

**Implementation Notes:**

- If the categorization file `aux_builds_categorization.csv` does not exist, the import process is aborted.
- Transforms image URLs to be browser-accessible.  
  Original example:  
  `https://storage.googleapis.com/download/storage/v1/b/buidlguidl-v3.appspot.com/o/builds%2F109d8bbf98675bc47528ea905.png?generation=1688403605357832&alt=media`  
  Modified example:  
  `https://storage.googleapis.com/buidlguidl-v3.appspot.com/builds/109d8bbf98675bc47528ea905.png`  
  > I made this transformation because the original URL forces a download when opened in the browser, and also behaves oddly when using "open image in new tab" on websites using the API URL.
- Maps build types and categories using both the API data and the categorization CSV. If a build type is available from BG3.5, we keep it; otherwise, we pull it from the CSV. For categories, we always use the CSV (since BG3.5 doesn’t include that field).
- Imports build–builder relationships and likes, ignoring builders that are not in the database (disabled builders or users with the same address but different casing).
- Cleans up orphaned builds that don’t have any builders. (We get all builds from the `/builds` endpoint, including those from disabled builders and duplicates with case-sensitive address issues.)

**Testing the Import**

My local categorization file for testing:  [aux_builds_categorization.csv](https://github.com/user-attachments/files/20094382/aux_builds_categorization.csv)

You can verify with the following builds:

- `bd119471-0ac3-5f72-9069-1ec83f402e5f` (ABI Ninja)
  - 1 main builder and 4 co-builders — both in origin and after import.
  - 36 likes in origin; 32 imported. The 4 missing are in the ignored builders list:  
    `0x1c80D2A677c4a7756cf7D00fbb1c1766321333c3`,  
    `0xa709FF766777C40bAC916558D9262eC4868496B4`,  
    `0x63D34D8188ad6D3CdbC1b1Cde031C93E47bdFd12`,  
    `0x3eaF68De4F9B8E80A3C0f54ec0983d062e778359`
- `3b9d71a3-5f0b-5975-bade-0728c7984e04` (DynOS 95)
  - 1 main builder and 3 co-builders — both in origin and after import.
  - 4 likes in origin and imported.

\* _For AI categorization, the only viable method I found was batching ~100 builds at a time using Gemini 2.5 Pro, which has a large context window. The results weren’t perfect but were better than basic Python keyword-based scripts.  
We could also consider using an API — if we’re okay with spending a bit of $$._


.gitignore
```diff
@@ -23,3 +23,5 @@ node_modules
 dist
 
 data/db
+
+packages/nextjs/services/database/bg-imports/aux_builds_categorization.csv
```
carletex, line 27: Can you move this to a more specific folder?

1. In nextjs folder (we already have a .gitignore there)
or
2. Create a .gitignore file in the bg-imports folder.

carletex (commented):
This iteration is looking cleaner!! Thanks Pablo.

Could you run the script against **Staging** Database? I imported the users & batches the other day.

Tagging @technophile-04 to do a pass! I'll do the same soon.

Thanks!

Pabl0cks:
> Could you run the script against Staging Database? I imported the users & batches the other day.

Done! In **staging** we have less ignored builders (only 5 because during my tests I did not have SRE import done), so: 
- `ABI Ninja` will show 35 likes (1 missing from `0x1c80D2A677c4a7756cf7D00fbb1c1766321333c3` who is one of the 5 missing users)

carletex:
Created https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/165 to facilitate testing.

.gitignore
technophile-04: .gitignore inside `bg-imports` dir sounds goood!

packages/nextjs/services/database/bg-imports/0003_build_data.ts
```diff
@@ -0,0 +1,212 @@
...
+      if (existingBuilders.has(address)) {
+        buildRelationshipsImported++;
+        return db.insert(buildBuilders).values({ buildId, userAddress: address, isOwner }).onConflictDoNothing();
+      }
+      ignoredBuilders.add(address);
+      return null;
+    }
+
+    // Upsert build-builders
+    const builderQueries = [];
```
technophile-04, line 156: nit: maybe call this buildBuilderQueries?

packages/nextjs/services/database/bg-imports/0003_build_data.ts
```diff
@@ -0,0 +1,212 @@
...
+      if (ownerRow) builderQueries.push(ownerRow);
+
+      // Add all co-builders
+      for (const addr of build.coBuilderAddresses) {
+        const coBuilderRow = getBuilderRow(build.uuid, addr, false);
+        if (coBuilderRow) builderQueries.push(coBuilderRow);
+      }
+    }
+    await db.executeQueries(builderQueries as any);
+    console.log(`✅ Build-builder relationships import complete (${buildRelationshipsImported} relationships created)`);
```
technophile-04, line 169: Ohh instead of maintaining a separate `buildRelationshipsImported` count, 

we could just do: `builderQueries.length` , same for `likeQueries`

technophile-04 (approved):
Tysm Pablo really nice job! Just went through the whole code and seems pretty intuitive and easy to understand and also seems good. 

> * For AI categorization, the only viable method I found was batching ~100 builds at a time using Gemini 2.5 Pro, which has a large context window. The results weren’t perfect but were better than basic Python keyword-based scripts.
We could also consider using an API — if we’re okay with spending a bit of $$.

Ohh I see it's PITA, so like when we do the final import to prod DB we would need to do this process again? Do you have the script or you did it manually pasting things to gemeni interface?

Maybe while importing to final DB if the BG builds counts grows a lot from current (1263)  and manual process is too tedious we can think of script? 

---

<details>

<summary>
    lol I always keep forgetting schema and feel underconfident while reviewing, I think asking chatGPT to create ER diagram from schema and looking to it and then trying to understand helps a lot, this is current status of our schema: 
</summary>

image


</details>

.gitignore
Pabl0cks: Created a `.gitignore` in `bg-imports` folder in https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/162/commits/36fa1fccd63afea5b7a2dfacfcb8f5fa3f0b0e7f 🙏

packages/nextjs/services/database/bg-imports/0003_build_data.ts
Pabl0cks: Updated variable names for both in 9115ee736140534904f59894e316aa5069533704 => `buildBuilderQueries` and `buildLikeQueries` 🙏

Pabl0cks:
Thx for the reviews! Just pushed the pending suggestions 🙌

> Ohh I see it's PITA, so like when we do the final import to prod DB we would need to do this process again? Do you have the script or you did it manually pasting things to gemeni interface?
> 
> Maybe while importing to final DB if the BG builds counts grows a lot from current (1263) and manual process is too tedious we can think of script?

I generated the data at 20250507_1237 (spain time), could generate the new builds created after this date and see. 

Gemini UI with 2.5 pro managed 100 builds per prompt fine (with more it would hallucinate). Probably will be able to add new ones in as single prompt (don't think we'll have more than 10-20)

ER diagram looks cool @technophile-04 ! Didn't know chatGPT was able to do that!! Last time I remember using a 3rd party tool which managed Neon connections, I think Neon should offer it as default!

technophile-04 (approved):
Tysm @Pabl0cks looking great and also working great 🙌

carletex (approved):
Great stuff, this is working great! Did some manual testing on staging, and everything seems on point.

Let's finish the rest of #113, and then we can run all the scripts in order (and in this case, generate the aux categorization file)

PR #164, opened 2025-05-08
builds crud

technophile-04:
### To test: 

You can checkout https://speedrunethereum-v2-git-builds-139-buidlguidldao.vercel.app/

### Description: 

We have added following things
1.  build submissions, 
2. updates 
3. deletions 
3. likes
4. and image upload (for build) 

#### 1. **API Endpoints for Builds**

- `POST /api/users/builds/submit`: Submit a new build (with EIP-712 signature validation).
- `PUT /api/users/builds/[buildId]/update`: Update an existing build (with signature and ownership checks).
- `DELETE /api/users/builds/[buildId]/delete`: Delete a build (with signature and ownership checks).
- `POST /api/users/builds/[buildId]/like`: Like/unlike a build (with signature validation).
- `POST /api/builds/upload-img`: Upload an image for a build (multipart/form-data, Firebase Storage).

#### 2. **Image Uploads**
- Images are uploaded via a dedicated endpoint to Firebase Storage.
- The code was taken from [1](https://github.com/scaffold-eth/buidlguidl-v3/blob/0d2a96dbf24c9b940c2c8cda706c6e8a4965495c/packages/backend/services/storage/storageFirebase.js#L9), [2](https://github.com/scaffold-eth/buidlguidl-v3/blob/0d2a96dbf24c9b940c2c8cda706c6e8a4965495c/packages/react-app/components/SubmitBuildModal.jsx#L57)
- Hence installed `react-dropzone` and `firebase` SDK

### Setup Notes

- Ensure Firebase credentials are set in your environment variables (see README).


technophile-04:
Just pushed somethings: 

- Fixed the toast (earlier it was showing below the modal) https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/164/commits/3481103e4546eccf59378d3dc14580ec4f9367d1
- instead of having download link get the normal publicUrl directly https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/164/commits/952aa20594988694d7791e4f81431ea3f63f9e2d  (pablo's mentioned about "Modified example" in https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/162#issue-3046091312 we directly get that url)

Pabl0cks (commented):
Nice job Shiv!! This was a big quest! 🙏

I've done a first round of tests at https://speedrunethereum-v2-git-builds-139-buidlguidldao.vercel.app/ (tried to set it up locally but must have done something wrong 😅) , here are my notes:

- Demo URL and Github URL throw format errors if you don't add https://https://. We could show a placeholder showing the format, and maybe also tweak the error notification a bit. Either that or let urls without http and we add it?
- You get Internal Server Error when you have spaces in the URL, we should show a informative notification so the user know where is the problem.
- Afraid if some people won't see the update button, then close the window, open and think it is stored (because of react state), but it's not really stored

| New | Old |
|---------|---------|
| image | image |

- When you add a builder not in SRE (vitalik.eth), we should throw a notification and not show it in the modal. When you try to update you get "Internal server error".
- Should we let add videos from other platforms? If not, i'd tweak the copy to Youtube video
- nit: - I'd set the category to NULL by default, if not all builds will be DeFi, For type field it's probably fine to set Dapp since a big % of builds will be Dapps, but maybe for DeFi we could have a "-- Select category --" or "-" option by default.
- nit: The "+" button in cobuilders is a bit small and may not look like a button. Increase the size and/or add a background with round shape?
- nit: Cobuilders should be able to edit the build (only to delete themselves)?
- nit: We could limit the title content shown to 2 lines
image

- nit: Should we hide the "View" button for now until the detail page is implemented?
- nit: Should we order categories dropdown alphabetically? For type I think it's fine like this.
- nit: Add a bit of spacing below the "+ New build" button, a bit too close to build cards.
image

- supernit: Should we add a background with "This builder doesn't have any builds", similar to challenges? (maybe not!)
image

- nit: Should we change "Demo URL" to "Live URL" (or other copy) if we want to store also production links?


Feel free to ignore all the nits, we could document the interesting ones in a new issue to tackle them in the future 🙌

packages/nextjs/app/builders/[address]/_components/builds/BuildFormModal.tsx
```diff
@@ -0,0 +1,340 @@
...
+                </button>
+              </div>
+              {form.coBuilders.map((address, idx) => (
+                <div key={idx} className="flex items-center gap-2 w-full">
+                  <div className="flex-1">
+                    <AddressInput
+                      placeholder="Builder Address"
+                      value={address}
+                      onChange={val => {
+                        console.log("val", val);
```
Comment (rin-st) [inline packages/nextjs/app/builders/[address]/_components/builds/BuildFormModal.tsx:253]: log

packages/nextjs/app/builders/[address]/_components/builds/BuildFormModal.tsx
```diff
@@ -0,0 +1,340 @@
...
+                        const updated = [...form.coBuilders];
+                        updated[idx] = val;
+                        setForm({ ...form, coBuilders: updated });
+                        console.log("form", form);
```
Comment (rin-st) [inline packages/nextjs/app/builders/[address]/_components/builds/BuildFormModal.tsx:257]: log

technophile-04:
Tysm Pablo! For great suggestion pushed some commits to handle most of them 🙌

> When you add a builder not in SRE (vitalik.eth), we should throw a notification and not show it in the modal. When you try to update you get "Internal server error".

Ohh yes :( not sure wha'ts the best way to handle this is but: 

1. We create a builder with role NULL and show register modal if the roll in null (currently we are not showing that modal if role is NULL) 
2. While submitting the form we make request to backend from FE checking if all the users exist and if not we show error that address is not present and can't add them as co builder 
3. We were talking about invite builder thing maybe somehow implement that? 

cc @carletex for this too 

> Afraid if some people won't see the update button, then close the window, open and think it is stored (because of react state), but it's not really stored

Umm the thing is that form is too big, maybe what we could do is have two fields in one line? Like in BGv3.5 also this happens but here we could have (type and category in one line) and (demo URL and github URL in online): 

| BGv3.5 overflowing | After | 
| ------- | ----- | 
| Screenshot 2025-05-15 at 3 21 49 PM | image |  

But again if co-builders list grow then we will have scroll bar (in update forms mostly) 

> nit: Cobuilders should be able to edit the build (only to delete themselves)?

Currently in BGv3.5 we are not allowing this too?

rin-st:
Remembered one more nit I noticed yesterday: after adding a build form is not cleared. So If you want to add second build and click "New build", you see first build data in the form

carletex:
> When you add a builder not in SRE (vitalik.eth), we should throw a notification and not show it in the modal. When you try to update you get "Internal server error".
> Ohh yes :( not sure wha'ts the best way to handle this is but:

Option 2 two seems best: next to the builder address we show a spinner while checking with the backend and then ✅ or ❌. But for this PR / first iteration, we could just check on the backend and ignore non-existent users (I don't like users being able to create other user hahah).

> nit: Cobuilders should be able to edit the build (only to delete themselves)?
> Currently in BGv3.5 we are not allowing this too?

I think only allowing the owner to edit is good enough for now!

packages/nextjs/app/api/users/builds/[buildId]/delete/route.ts
```diff
@@ -0,0 +1,37 @@
...
+      signature,
+      buildId: buildId,
+    });
+
+    if (!isValidSignature) {
+      return NextResponse.json({ error: "Invalid signature" }, { status: 401 });
+    }
+
+    const isOwner = await isOwnerOfBuild(buildId, address);
+    if (!isOwner) {
```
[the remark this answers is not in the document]
Comment (technophile-04) [reply packages/nextjs/app/api/users/builds/[buildId]/delete/route.ts]: fixed at https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/164/commits/70a7b5513e1a54237c97e21bf7146e87f9713ba3 and https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/164/commits/3dc068ab1964c10e6007ac20cb647ca6a103a76c

packages/nextjs/app/api/users/builds/submit/route.tsx
```diff
@@ -0,0 +1,40 @@
...
+export type SubmitBuildPayload = {
+  address: string;
+  signature: `0x${string}`;
+} & BuildFormInputs;
+
+export async function POST(request: NextRequest) {
+  try {
+    const { address, signature, ...build } = (await request.json()) as SubmitBuildPayload;
+
+    if (!address || !signature || !build.name) {
```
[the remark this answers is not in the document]
technophile-04: Oh yes made desc to deafult, removed gh repo URL to required because of https://github.com/BuidlGuidl/SpeedRunEthereum-v2/commit/78e5773cd2fc0f76393d443b420f613be5a48d24#r157167049

packages/nextjs/services/database/repositories/builds.ts
```diff
@@ -0,0 +1,179 @@
...
+export type BuildLike = InferSelectModel<typeof buildLikes>;
+
+export type BuildInsert = InferInsertModel<typeof builds> &
+  Omit<InferInsertModel<typeof buildBuilders>, "buildId"> & {
+    coBuilders?: string[];
+  };
+
+export const createBuild = async (build: BuildInsert) => {
+  // Insert the build
+  const [insertdBuild] = await db.insert(builds).values(build).returning();
```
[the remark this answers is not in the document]
technophile-04: Oops, thanks fixed it https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/164/commits/84999a10c5a0862a0945a25d7f0d1bf6c9dc49f9

technophile-04:
> Option 2 two seems best: next to the builder address we show a spinner while checking with the backend and then ✅ or ❌. But for this PR / first iteration, we could just check on the backend and ignore non-existent users (I don't like users being able to create other user hahah).

Makes sense! For this PR we just remove the builder address which are not present in user table https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/164/commits/8805f770f35d10fa46fba4be7de275eeba7a249c


Also added `HIDE_BUILDS` var discussing with Carlos so that we hide builds for initial 1 week after merging until we cut off the link b/w BG and SRE  

Thanks all for the reviews and suggestions 🙌

packages/nextjs/app/api/users/builds/submit/route.tsx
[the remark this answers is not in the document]
rin-st: nit: even if it's design, we need possibility to see it, so we need one of demoUrl, videoUrl, githubUrl, (or imageUrl if it will be enough, but I'm not sure about it). Not sure how to set the info on the frontend though, maybe just plain text or asterisk on all that fields with tooltip

Pabl0cks:
TYSM for the quick improvements @technophile-04  🔥 It's looking great 💪

On build submissions, we are clearing the data too soon now. Just after you hit submit it gets cleared, so if you cancel the signing, or have some problem, you have to type it all again.

data-clearing-submission

> Umm the thing is that form is too big, maybe what we could do is have two fields in one line? Like in BGv3.5 also this happens but here we could have (type and category in one line) and (demo URL and github URL in online):

Maybe 2 URL fields in one line is too packed. Type & category could be an option (if they don't look too packed) and if it makes the "Submit" to show on most desktop screens. If not, maybe we can leave it like this and see if people is reporting some UX problems here.

We could test a sticky modal header if people are having UX problems, or it would look too weird?


Thanks again Shiv! ♥

technophile-04:
> Maybe 2 URL fields in one line is too packed. Type & category could be an option (if they don't look too packed) and if it makes the "Submit" to show on most desktop screens. If not, maybe we can leave it like this and see if people is reporting some UX problems here.

Just moved Type and category in one line, but if we don't like we could have them on line...lol it's hard to show that submit / update button without scroll 

>  sticky modal header if people are having UX problems, or it would look too weird?

Ohh you mean to say we make this section sticky? 

Screenshot 2025-05-16 at 2 09 49 PM

But even if we do that there would still be scroll no?  

Thanks Pablo! 🙌

packages/nextjs/app/builders/[address]/_components/builds/SubmitNewBuildBtn.tsx
```diff
@@ -0,0 +1,31 @@
+"use client";
+
+import { useRef } from "react";
+import { BuildFormModal } from "./BuildFormModal";
+import { PlusIcon } from "@heroicons/react/24/outline";
+import { useSubmitBuild } from "~~/hooks/useSubmitBuild";
+
+export const SubmitNewBuildBtn = () => {
```
Comment (carletex) [inline packages/nextjs/app/builders/[address]/_components/builds/SubmitNewBuildBtn.tsx:8]: Let's call this `SubmitNewBuildButton`

packages/nextjs/app/builders/[address]/page.tsx
```diff
@@ -76,14 +83,44 @@ export default async function BuilderPage({ params }: { params: { address: strin
...
+                  This builder hasn&apos;t completed any challenges.
+                </div>
+              )}
+            </div>
+            {/* Builds */}
+            {!HIDE_BUILDS && (
+              <div className="w-full">
+                <div className="flex justify-between items-center">
+                  <h2 className="text-2xl font-bold mb-0 text-neutral pb-4">Builds</h2>
+                  <SubmitNewBuildBtn />
```
Comment (carletex) [inline packages/nextjs/app/builders/[address]/page.tsx:104]: This button should only show for the profile owner. Right now, it's showing on every profile.

Also consider moving all of this section into a component (since this profile page it's going to get bigger and bigger)

packages/nextjs/app/builders/[address]/_components/builds/SubmitNewBuildBtn.tsx
Comment (carletex) [reply packages/nextjs/app/builders/[address]/_components/builds/SubmitNewBuildBtn.tsx:8]: oh, maybe we can tweak the other `Btn` too? It's kind of a nit, but I don't see the benefit of using `Btn` (only that it has fewer letters?)

packages/nextjs/app/builders/[address]/_components/builds/LikeBuildBtn.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const { likeBuildMutation, isPending } = useBuildLike({
+    onSuccess: () => {
+      notification.success(`Build ${isLiked ? "unliked" : "liked"} successfully!`);
+    },
+  });
+
+  return (
+    <button onClick={() => likeBuildMutation({ buildId, action: isLiked ? "unlike" : "like" })} disabled={isPending}>
+      {isLiked ? <HeartIconSolid className="h-5 w-5 text-red-500" /> : <HeartIconOutline className="h-5 w-5" />}
+    </button>
```
Comment (carletex) [inline packages/nextjs/app/builders/[address]/_components/builds/LikeBuildBtn.tsx:27]: nit: could we make a hover effect here? If not you only get the "cursor pointer"

packages/nextjs/app/builders/[address]/_components/builds/SubmitNewBuildBtn.tsx
Comment (technophile-04) [reply packages/nextjs/app/builders/[address]/_components/builds/SubmitNewBuildBtn.tsx]: Renamed at https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/164/commits/c68f9b66dfb0ce82caf88acf3b3aef31e40fc4ec 🙌

packages/nextjs/app/builders/[address]/_components/builds/LikeBuildBtn.tsx
Comment (technophile-04) [reply packages/nextjs/app/builders/[address]/_components/builds/LikeBuildBtn.tsx]: Used GPT to add it https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/164/commits/d324127f412b9990a019bca59ca447fd9afe147a not sure if it's too much

carletex (approved):
Good stuff, Shiv! This was a big one. Thanks Pablo and Rinat for the thorough reviews.

There might be some details to polish, but we can iterate later. Merging this!!

PR #166, opened 2025-05-09
Admin submenu

rin-st:
Idea for this menu https://github.com/BuidlGuidl/SpeedRunEthereum-v2/issues/151#issuecomment-2863425858

Menu for admins:
Screenshot 2025-05-09 at 13 22 41

For regular users menu is the same as before

Also
- added admin badge
- moved activity page to `/admin/activity` and added `isAdmin` to get activity api
- fixed Edit button show condition on builder page, so now it doesn't require siwe and works for users with admin role

Fixes #151

PR #167, opened 2025-05-10
Always siwe admin

rin-st:
Reasons for the changes
- people can use impersonator so they can see content which relies just on `user.role === UserRole.Admin`
- it almost doesnt make sense to login as an admin without having access to admin functionality

Changed siwe behavior, so
1. User logs in. If he has Admin role then he requested for siwe. If user is not the admin, behavior will be the same.
2. I removed sign in for `/admin/**` pages. So since admins always do siwe, pages just accessible. For other users - access denied
3. All client behavior except [1] now depends on `isAdmin` from `useAuthSession`. 

Also, if someone will try to use impersonator with admin address
- because of [1] he will not even pass Sign in page. And since that page is always shown for admin he will not see other pages too
- even if we disable [1], because of [3] all the admin features will be disabled until siwe which is not possible with impersonator

---
notes:
- PR to `admin-menu` branch ( #166 )
- first commit changes siwe behavior, second commit adds changes to use `isAdmin`  from `useAuthSession`

### DIFF
--- a/packages/nextjs/app/_components/AdminSiwe.tsx
+++ b/packages/nextjs/app/_components/AdminSiwe.tsx
@@ -1,0 +1,18 @@
+"use client";
+
+import { SignIn } from "../admin/_components/SignIn";
+import { Session } from "next-auth";
+import { useAccount } from "wagmi";
+import { useUser } from "~~/hooks/useUser";
+import { UserRole } from "~~/services/database/config/types";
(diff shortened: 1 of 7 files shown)


rin-st:
Closing this for now since write operations are already protected and because of that hiding data for impersonator makes not so much sense

PR #170, opened 2025-05-16
create build details page

technophile-04:
### Description: 

Example url: https://speedrunethereum-v2-git-build-details-page-buidlguidldao.vercel.app/builds/bd119471-0ac3-5f72-9069-1ec83f402e5f

Updated this branch to use same ENVS as builds-139 branch


packages/nextjs/services/github.ts
technophile-04, line 1: This code was take from BGv3.5. 

Actually earlier I tried updating `fetchGithubReadme` to handle full github URL and it to return the `raw.githubusercontent.com` URL and then fetch README. 

But there are cases where a user might have transferred the repo or something and in that case the simply appending org/user name and repo doesn't work like github doesn't redirect to correct URL and hence we use the `api.github.com` in that case which resolve its nicely. 

And example of this SRE build which had : 

https://github.com/moonshotcollective/scaffold-directory 

And if you: 

https://raw.githubusercontent.com/moonshotcollective/scaffold-directory/main/README.md it doesn't resolve anything

carletex:
Closing this: https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/172

PR #171, opened 2025-05-16
Onboarding batches banner

### LINKED ISSUE #116: Update Join BG Banner
People won't join the BG anymore so we should remove / replace the banner (unlock your profile, join a batch... TBD)

This will be the last piece of #113 (and it might be part of #115).

+ Remove "Builder has upgraded to BG"

rin-st:
Batches banner and functionality described in #116 

Screenshot 2025-05-16 at 14 13 39

Also in scope of this pr 
- removed socials requirement as it was when joining bg since user joining tg channel. Or do we need it?
- updating batch was a bit broken, fixed it

\+ see comments

Note: Currently I'm adding users to latest open batch if it exists. But as I understand in the future there could be parallel multiple batches (universities etc). If I'm right, behavior should be changed, for example add flag for an open batch where to add users.

Fixes #116


packages/nextjs/utils/date.ts
```diff
@@ -1,22 +1,51 @@
...
+
+  return rtf.format(diffMinutes, "minute");
+};
+
+export const getRelativeTime = (timestamp: string | number | Date) => {
+  const { diffMs } = getTimeDiff(timestamp);
+
+  if (diffMs < 0) return getPastRelativeTime(timestamp);
+
+  return getFutureRelativeTime(timestamp);
```
rin-st, line 48: Returning a bit different formats for past and future timestamps here. I think it's fine for future dates to give more detailed view since user kindof waits for that time. And for now we use it just in one place. Or better create separate function for detailed relative time?

packages/nextjs/app/api/users/join-batch/route.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (!isValidSignature) {
+      return NextResponse.json({ error: "Invalid signature" }, { status: 401 });
+    }
+
+    const updatedUser = await updateUser(user.userAddress, {
+      batchId: latestOpenBatch.id,
+      batchStatus: BatchUserStatus.CANDIDATE,
+    });
+
+    waitUntil(trackPlausibleEvent(PlausibleEvent.JOIN_BATCH, {}, req));
```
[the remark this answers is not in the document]
carletex: Just added the `joinBatch` goal on Plausible.

packages/nextjs/utils/date.ts
carletex: I think this is cool.

thanks!

Pabl0cks (approved):
Good job @rin-st !! Is working nicely.

Nitpick: when you close the current batch and create a new one (I did it on the database), the latestOpenBatch remains the old one, not sure if we should force the refresh either when we update the batches from the UI or directly in the hook that fetches the info from database. A bit edge case probably.

Have tested the whole SRE workflow with a fresh user and is working well 🙏

carletex (approved):

PR #172, opened 2025-05-16
Build detail page (rebase/cherry pick)

carletex:
Since https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/170 was branched from #164 it was a bit messy so cherry picked the commits and created this PR.


carletex:
This is working great @technophile-04 

I just pushed some nitpicks, but this is ready to go!

PS. Not worrying too much about styling since we'll do a full pass with Andrea / Hunter.

PR #173, opened 2025-05-20
Disallow /builds/ in robots.txt to block crawlers from build pages

Pabl0cks:
Builders pages are currently disallowed (it's where build pages are currently linked) and we have no /builds page yet, but adding it to robots just in case.

This way we'll also be ready for whenever we open the /builds page.


carletex (approved):
Thanks Pablo!

PR #180, opened 2025-05-29
Ens with refresh button

rin-st:
From https://github.com/BuidlGuidl/SpeedRunEthereum-v2/issues/45#issue-2914598028

> c. Have a "refresh" button in the user profile, so they can click it and we'll fetch it and cache it.
> c might be good (it's the most efficient at least haha), especially if we have the ENS stuff as a side quest (get your ENS! Read this article on what ENS is and how to get One... etc)

Possible implementation of refresh ENS button. But it's very weird to click to refresh ENS when you already have that ENS in some places on the page.
https://github.com/user-attachments/assets/8673cf56-a746-4b4b-98fd-1dfa228d0eda

If user created ENS right now and it's not yet resolved on ENS side then there will be no ENS on the page, but button also most likely won't work

### DIFF
--- a/packages/nextjs/app/api/users/[address]/update-ens/route.ts
+++ b/packages/nextjs/app/api/users/[address]/update-ens/route.ts
@@ -1,0 +1,50 @@
+import { NextResponse } from "next/server";
+import { isUserAdmin, updateUser } from "~~/services/database/repositories/users";
+import { isValidEIP712UpdateEnsSignature } from "~~/services/eip712/ens";
+import { publicClient } from "~~/utils/ens-or-address";
+
+type UpdateEnsPayload = {
+  address: string;
+  signature: `0x${string}`;
+};
+
+export async function PUT(req: Request, { params }: { params: { address: string } }) {
+  try {
+    const { address, signature } = (await req.json()) as UpdateEnsPayload;
+
+    if (!address || !signature) {
+      return NextResponse.json({ error: "Address and signature are required" }, { status: 400 });
+    }
+
+    const isAdmin = await isUserAdmin(address);
+    const addressToUpdateEns = params.address.toLowerCase();
+
+    if (address.toLowerCase() !== addressToUpdateEns.toLowerCase() && !isAdmin) {
+      return NextResponse.json({ error: "Address mismatch" }, { status: 400 });
+    }
+
+    const isValidSignature = await isValidEIP712UpdateEnsSignature({ address, signature });
+
+    if (!isValidSignature) {
+      return NextResponse.json({ error: "Invalid signature" }, { status: 401 });
+    }
+
+    try {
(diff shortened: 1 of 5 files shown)


packages/nextjs/components/scaffold-eth/Address/AddressWithCachedEns.tsx
technophile-04, line 1: I think we are not using this anywhere

technophile-04 (commented):
Thanks @rin-st, this is looking great! Just added a couple of small comments but looking overall good!

rin-st:
> Ohh just while using publicClient in server https://github.com/BuidlGuidl/SpeedRunEthereum-v2/blob/main/packages/nextjs/utils/ens-or-address.ts we should vigilant because it Alchemy seems to block the request from server saying :
> If we have allowlist in prod set

maybe we can just add it to allowlist or use bg rpc?

packages/nextjs/components/scaffold-eth/Address/AddressWithCachedEns.tsx
rin-st: yes, see initial comment

> added AddressWithCachedEns component which uses useUser to get ENS. For cases where you have address and don't have ENS. Could be used instead of Address component on blockexplorer page, debug page, Faucet component etc but since we don't need any of that, component AddressWithCachedEns also is not in use

I created it just in case we don't have user data near the address component

technophile-04: Ohh my bad didn't read it, I would probably remove this since I think we won't using debug, facuet etc component and we can add this back if we plan to to them so 🙌

--- 

Initially I was thinking is since we are sure about https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/180#discussion_r2140440968 what if we remove this component completely. 

and in `Address` component instead of passing the the prop `cachedENS` we do the same logic as here. 

But then while rendering something like bulk users like in builders table even thought we are getting users ens in array, we would still be make each user request in `Address` component soo not worth it. 

Soo yeah let's just remove this component I feel and keep things as it 🙌

technophile-04:
> maybe we can just add it to allowlist or use bg rpc? 

Ohh I was saying we have to do this https://github.com/BuidlGuidl/SpeedRunEthereum-v2/blob/main/packages/nextjs/utils/ens-or-address.ts if we have configured allowlist for prod alchemy key. Liken even if we allow the speedrunethereum.com domain in allowlist the backend won't be able to make the request until and unless it itself set header origin as speedrunethereum.com .

Just now checked the prod alchemy key configuration which we are using in SRE and we are allowing all domains their soo it should be fine and we don't need to do anything 🙌

carletex:
Not super happy with the result, but at least it's more subtle haha


image
image

PR #186, opened 2025-06-05
search on builders page

### LINKED ISSUE #183: Search on /builders
As in https://app.buidlguidl.com/builders, let's have a search for the builders page:

- Search in the ens name (when #179 is ready!) + socials.
- 3 chars minimum to start querying database
- Debounce

It can be added at the top of the page, below the Builder count. Just show a spinner in the table when the query is happening and show results or "no builders found" text.

technophile-04:
### Description: 

As described in #183 it works similar: 

> - Search in the ens name (when #179 is ready!) + socials.
> - 3 chars minimum to start querying database
> - Debounce

> It can be added at the top of the page, below the Builder count. Just show a spinner in the table when the query is happening and show results or "no builders found" text.

Instead of showing spinner showing the similar UI as "Fetching data" but happy to update it to show spinner 🙌 For no data we show : 

image

Fixes #183

### DIFF
--- a/packages/nextjs/app/api/users/sorted/route.ts
+++ b/packages/nextjs/app/api/users/sorted/route.ts
@@ -9,8 +9,9 @@ export async function GET(request: NextRequest) {
     const start = parseInt(searchParams.get("start") ?? "0");
     const size = parseInt(searchParams.get("size") ?? "0");
     const sorting = JSON.parse(searchParams.get("sorting") ?? "[]");
+    const filter = searchParams.get("filter");
 
-    const data = await getSortedUsersWithChallengesInfo(start, size, sorting);
+    const data = await getSortedUsersWithChallengesInfo(start, size, sorting, filter || "");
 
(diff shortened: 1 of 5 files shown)


carletex (approved):
This is looking great, Shiv! Thank you.

I pushed some little tweaks.

The "fetching data" at the bottom bugs me a bit. I guess that something that shows in the search input / next to the input will be better (more visible). But I also get that we are reusing the `isFetching` / `queryFn`. So this is more that enough.

Merging!

carletex:
> We could mention in search placeholdr but it was getting hidden also maybe it's too verbose and this looks fine?
> As an option we could add smaller text "at least 3 characters" below the input when there are only 0-2.
> Or "Search by ENS or socials" before the input and "at least 3 characters" as a placeholder

I get this, but also I think it's not a big deal... so simpler better for now (also, it's probably admins who will use that a lot... at least for now!)

PR #188, opened 2025-06-06
Add All Builds Page

ChangoMan:
Adding a page that shows all the builds, with text search and filtering on category and type. I had made a previous version that fetched the data on the server side, but the text search wasn't ideal.

This is the client side version and it seems more aligned with the other pages that have text filtering.

Screenshot 2025-06-06 at 10 10 57 AM


packages/nextjs/app/builds/page.tsx
```diff
@@ -0,0 +1,16 @@
...
+export const metadata = getMetadata({
+  title: "All Builds",
+  description: "View all the builds by the Speed Run Ethereum community",
+});
+
+export default async function AllBuildsPage({
+  searchParams,
+}: {
+  searchParams: { category?: BuildCategory; type?: BuildType };
+}) {
```
ChangoMan, line 14: Note: I am using search params from the server side page because the build was failing when I had this whole page under `"use client"`. It was related to https://nextjs.org/docs/messages/missing-suspense-with-csr-bailout

carletex:
Thanks for this, Hunter! This is looking pretty good.

I haven't reviewed deeply, but commenting on a few things:

- Let's sort them by likes DESC (check `build_likes` table) 
- Show the likes on the card  + allow like/unlike (I think you can copy from the builds card inside builders portolio page)
- Align the view button at the bottom of the card

> I had made a previous version that fetched the data on the server side, but the text search wasn't ideal.
> This is the client side version and it seems more aligned with the other pages that have text filtering.

I think we have a good balance in the /builders page. What I'd do here:

Let's fetch only the first 30/50 from the backend (or maybe 32... 48 or any multiple of 4, since you are showing 4 per row) items
  - This will make us move the text search to the backend again. Agree that text search is amazing when searching in memory/client, but we already have +1k builds, and if it keeps growing it might get clunky.
  - Could we do the infinite scroll thing here too?

---

Also, feel free to leave some items for future PRs. Happy to merge a v1 and iterate.

Thanks!!!

ChangoMan:
@carletex Thanks for the feedback, I've added the sorting by likes and the like button to the cards. I've also fixed the view button alignment and limited the results to `48`.

The text search should still work, I am passing the value all the way to the backend for the text search. I can look into the infinite scrolling in a future PR

carletex:
Looking great!! Thanks, Hunter

I think this is perfect for the first iteration. Tagging @technophile-04 so he can do a deeper code review (no rush, Shiv!)

> I can look into the infinite scrolling in a future PR

Great! It'd also be nice to show the count of all the builds (filtered or not) at the top

packages/nextjs/app/builders/[address]/page.tsx
```diff
@@ -25,7 +25,7 @@ export async function generateMetadata({ params }: Props): Promise<Metadata> {
   const { ensName, shortAddress } = await getEnsOrAddress(address);
 
   // Default title and description
-  const title = `${ensName || shortAddress} | Speed Run Ethereum`;
+  const title = `${ensName || shortAddress}`;
```
Comment (ChangoMan) [inline packages/nextjs/app/builders/[address]/page.tsx:28]: This was a lingering bug from when I did the builder og image a while ago

packages/nextjs/services/database/repositories/builds.ts
```diff
@@ -197,3 +198,28 @@ export const getBuildByBuildId = async (buildId: string) => {
...
+    where: and(
+      category ? eq(builds.buildCategory, category) : undefined,
+      type ? eq(builds.buildType, type) : undefined,
+      nameSearch ? ilike(builds.name, `%${nameSearch}%`) : undefined,
+    ),
+    with: {
+      likes: true,
+    },
+    orderBy: (builds, { desc, sql }) => [desc(sql`(SELECT COUNT(*) FROM build_likes WHERE build_id = ${builds.id})`)],
+    limit: 48,
```
technophile-04, line 221: Instead of fetching all the builds we just limiting to top 48 builds, shall we have infinite scroll behavior on this page? Maybe we can tackle this in other PR? 

cc @carletex

PR #192, opened 2025-06-10
Add All Guides page

Pabl0cks:
Introduces a `/guides` page listing all the guides in a card grid.

- Includes a client-side search bar (with icon) for filtering by title or description.
- Styling and UX trying to match the /builds and /builders pages.
- Guides are fetched server-side for SEO, with fast client-side filtering.


carletex (approved):
Thanks Pablo! Looking good.

Pushed a few commits:
- Extract function to get all guides to the already created service
- Use the existing Guide type
- Remove revalidate (not needed since when we modify / add  / remove a guide we need to merge a PR => will rebuild the site)
- Fix weird classname formatting

carletex:
Fixed! They were some absolute positioning that it shouldn't be there. 

(I'm sure that there are also some classes that are not needed, but maybe not worth it to go one by one)

Merging, thanks!!

PR #195, opened 2025-06-11
Update ens avatars scripts

rin-st:
Used `text` in schema since some people use svg for their avatars and some of them are more than 60k characters.

Script is slower than caching ens so it can take some time

Tested on staging and it works fine

Not showing it for now on portfolio page. Can add it to #180 after merging this or make a separate PR later

This PR has some duplicating like chunk function but I think it's fine, can change if needed


packages/nextjs/services/database/scripts/update-ens-avatars.ts
```diff
@@ -0,0 +1,118 @@
...
+    });
+
+    // Add delay between batches to avoid rate limiting
+    await new Promise(resolve => setTimeout(resolve, 200));
+  }
+
+  return results;
+}
+
+async function updateEnsAvatars() {
```
carletex, line 67: Question: If we wanted to use the multicall, I guess we should dissect how viem does it internally, right? We half-do it on ENS names, right?

e.g. https://github.com/wevm/viem/blob/main/src/actions/ens/getEnsText.ts#L118

image

---

Maybe it's not worth it for now. But if we want to run a cron script every week or something, I think we can benefit if we can multicall 500 request into one.

rin-st: > Question: If we wanted to use the multicall, I guess we should dissect how viem does it internally, right? 

Yes, updated the function using viem implementation, so now it's getting ens avatars records in batches of 500, so we make much less RPC calls for it :). 

But it returns pure data, it could be ipfs or ipns, nft (encoded or not), regular image link etc so we need to convert it using [parseAvatarRecord](https://github.com/wevm/viem/blob/main/src/actions/ens/getEnsAvatar.ts#L88) anyway, which is slow since it does another rpc calls for nfts, check if image urls are valid, encode/decode data etc. So regarding the script time it's not so much better, local tests are 125 sec old vs 77 sec new version.

---

> We half-do it on ENS names, right?

Partially, just to understand which addresses has ens names, but not for getting that ens. I didn't use internal viem implementation. If we do it like in this script it could be much more effective and faster (500x?) since it doesnt need `parseAvatarRecord`

carletex: > So regarding the script time it's not so much better, local tests are 125 sec old vs 77 sec new version.

Well, still a good improvment!

rin-st:
> Hey Rinat, tested on staging, and it picked up some avatars but not all?
> Mine is NULL:

Weird, I tested it yesterday and especially with your ens and it worked. Probably I updated something after.

Looks like batch size was too large and because of that, there were too many `parseAvatarRecord` in parallel, something like 1/4 of batch size, ~100-150. Changed batch size to 250 and now Ens updates for 697 users instead of 690 (you were in that not updated 1% lol), and your ens resolves too. But script became slower

> OH, let's also add it to the register route? (as we did with the ens name). If the registered user already have an avatar, let's cache it

added it, thanks [e8b6218](https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/195/commits/e8b6218b7ddf3845f4cab941487607a5886b98b6)

PR #197, opened 2025-06-12
Fix update ens multicall

rin-st:
Fixed Update ens script so now it works using multicall with the same functionality as in viem. It allowed to get rid of unnecessary steps like finding resolvers and `publicClient.getEnsName()` for every address

cc @technophile-04


carletex (approved):
I think this is great. This script is meant for 1 use only, but we can reuse it if we set up some crons

PR #199, opened 2025-06-12
Improve Filters on Builds Page

ChangoMan:
- Moves the filters to the top of the page
- Adjust filter size
- Add number to indicate number of builds showing
- Add indicator when a Filter is active

Screenshot 2025-06-13 at 7 29 08 AM


carletex (approved):
Great iteration. Thanks Hunter!

PR #200, opened 2025-06-13
Add OCL challenge

carletex:
- Removing the Stablecoins challenge for now.
- Added the challenge to the database in staging and production (sortOrder = 5 so it matches the autograder)

image

Database data:

image


escottalexander (approved):
LGTM!

PR #202, opened 2025-06-16
Revalidate path api

rin-st:
Adds the possibility to revalidate path on demand

We have
```
// 6 hours
export const revalidate = 21600;
```
and can have similar on other pages in future. So if we want to make a hotfix to the challenge Readme, we need to wait up to 6 hours to update, after that hotfix PR is merged. (in my case, it was updated in less than 1 hour, though). This PR fixes that behavior. So how to use:

add `REVALIDATION_TOKEN` to env vars, and after that, we can revalidate path whenever we want

```
curl -X POST "http://localhost:3000/api/revalidate?token=<refalidate-token>&path=/challenge/<challenge-id>"
```

What do you think? Maybe there's another way to revalidate paths I don't know?

Note: I didn't test it since it looks like there's nothing to break, it's not so quick, and not even sure you agree to add this feature. Can test if needed later

Docs: https://nextjs.org/docs/app/api-reference/functions/revalidatePath


carletex (approved):
Not sure how much we will use this, but since it's a small thing, happy to merge!

(In general, I prefer to add stuff to the codebase the moment we have a need for it)

PS. I generated a REVALIDATION_TOKEN and added it to Vercel.

Thanks Rinat!!

PR #204, opened 2025-06-17
Ocl image replacement

andrealbiac:
Quick PR to fix the current image, in which you could see this horizontal lines from the SVG rendering:

Captura de pantalla 2025-06-17 a las 10 54 18

I keep forgetting this happens until I see it live (it doesn't show in the Figma) and then I can fix it by merging the color blocks.

*I can't see the local deployment for some reason so will need to check on Vercel that it's replaced correctly, thanks!


carletex (approved):
Thanks Andrea!

PR #208, opened 2025-06-19
set prediction markets challenge live

phipsae:
Hi all, 

@technophile-04 already will make the changes to the database (e.g. github, autograder to true), thanks!
I've just removed the "Coming Soon" label in the UI.

The autograder should already be active: https://github.com/austintgriffith/speedrun-grader/pull/16 

From my side, everything looks good to go. Pls give the challenge one final check.
https://github.com/scaffold-eth/se-2-challenges/tree/challenge-prediction-markets 

Thanks!


technophile-04 (approved):
Thanks @phipsae! Just updated prod databse with autograding, dependency and repo URL also thanks to @Pabl0cks  🙌!

PR #210, opened 2025-06-26
Landing page visual tweaks

andrealbiac:
Added some tweaks to the page:

- a slide in animation for the 'Already comfortable building...' modal. (lmk how you feel about this) - only in desktop. (vibe coded the hook so I hope it didn't do anything weird)
- hover effect for the 'Start Challenge 0' button
- parallax effect on scroll for the background clouds
- made the intro text centered in desktop and mobile - the rest are left-aligned but this one makes sense to me centered, since it's centered with the logo
- adjusted the Hero logo to a darker green to add some contrast
- deleted the header logo in desktop to avoid showing it twice at the same time (header + hero logos)
- adjusted some paddings and margins
- tweaked some dark mode elements too, since the hero white bg color we never used before so it didn't had a dark mode version of it

*I also had to create a new layout file to handle the metadata errors that I was getting, not sure why I was getting them. Maybe there's another way to do this (that was the AI fix but I read it's fine to have multiple layout files?)

Captura de pantalla 2025-06-26 a las 14 38 58
Captura de pantalla 2025-06-26 a las 14 38 53

Captura de pantalla 2025-06-26 a las 14 39 07

Lmk what you think or if I should change anything! Tagging @ChangoMan in case he wants to take a look as well

### DIFF
--- a/packages/nextjs/app/start/_components/AnimatedWindow.tsx
+++ b/packages/nextjs/app/start/_components/AnimatedWindow.tsx
@@ -8,11 +8,11 @@ interface AnimatedWindowProps {
 }
 
 export const AnimatedWindow = ({ className = "", children }: AnimatedWindowProps) => {
-  const windowSlideStyle = useSlideInFromRight(0.05, 350);
+  const windowSlideStyle = useSlideInFromRight(0.08, 300);
 
   return (
     <div
-      className={`mb-10 mx-auto w-80 h-40 bg-[url('/assets/start/window-1.svg')] bg-center bg-no-repeat lg:mb-0 lg:absolute lg:right-[5%] lg:-bottom-20 ${className}`}
+      className={`mb-10 mx-auto w-80 h-40 bg-[url('/assets/start/window-1.svg')] bg-center bg-no-repeat lg:mb-0 lg:absolute lg:right-[15%] lg:-bottom-20 ${className}`}
       style={windowSlideStyle}
     >
       {children}


ChangoMan:
@andrealbiac Regarding this:

> I also had to create a new layout file to handle the metadata errors that I was getting, not sure why I was getting them. Maybe there's another way to do this (that was the AI fix but I read it's fine to have multiple layout files?)

The reason this happened is because in Next.js, there are two types of files. There are "server" files and "client" files (which means it can interact with the browser).

The `getMetadata` stuff only works on "server" files. But the parallax and scrolling hooks that were added only work on "client" files. So the AI added the "use client" to the top of the `page.tsx` which makes the whole page into a "client" file, making the metadata stuff not work anymore.

Having multiple layout files is indeed okay in Next.js but the best practice is having as many "server" files as you can. It's also not best practice to convert the entire `page.tsx` into a "client" file.

An alternative solution would be to take the "use client" out of `page.tsx`, changing it back to a "server" file. Then, take the clouds and window, and extract them out into a different component. Then, in the new components, you can add the `useParallax` hooks and add the "use client" only in those component files. You can then delete that new `layout` file and put the `getMetadata` back into `page.tsx`

Let me know if this makes sense or if there are any questions!

packages/nextjs/app/start/page.tsx
```diff
@@ -57,20 +63,23 @@ const StartLandingPage = () => {
             </p>
           </div>
         </div>
-        <div className="mb-8 mx-auto w-80 h-40 bg-[url('/assets/start/window-1.svg')] bg-center bg-no-repeat lg:mb-0 lg:absolute lg:right-[5%] lg:-bottom-20">
+        <div
+          className="mb-10 mx-auto w-80 h-40 bg-[url('/assets/start/window-1.svg')] bg-center bg-no-repeat lg:mb-0 lg:absolute lg:right-[5%] lg:-bottom-20"
+          style={windowSlideStyle}
+        >
```
ChangoMan, line 69: This slide in effect is pretty cool, but there are more modern ways to animate things. It would include adding a new library and it's harder to use, but the results can look a lot better.

Take a look at https://examples.motion.dev/ for some examples

andrealbiac:
Okay got it, this is super helpful cause I had some back and forth with the AI with this but wasn't aware of the client/server distinction

I'm gonna do that then, turn it back into a server file and taking the animations out to separate components, will come back with questions if I can't make it work

Thanks so much! @ChangoMan

ChangoMan:
> I think I fixed it, can you take a look a it please @ChangoMan ? I did it with the AI but followed your instructions and I think it looks good
> 
> Deleted the `'use client'` and move the clouds and slide in window out of `page.tsx`, and deleted the `layout` page as well.

Yeah, looks good to me! Nice work. I would like @carletex to take a quick look and test if possible, since he had an issue with scrolling detection on his machine.

carletex:
I'm the guy with the scroll issue haha

I don't see the parallax on the clouds. The only thing that's moving for me is this:

image

What do you guys think about make it a bit more subtle? I like that it moves (it draws attetion) but I fell that you have to scroll a lot to see the whole thing, and it might defeat the purpose of this "shortcut CTA"

andrealbiac:
Delayed the window animation so you see it early on while scrolling. Trying to clean up opened PRs on this to focus on next improvements. How does it look for you now @carletex ?

carletex:
Parallax doesn't work for me :/ The slide from the right works tho.

The animations are icing on the cake, so I'd focus on the _big_ improvements next (showcasing all the challenges in the map, think about the playground, tweak copy if needed).

So maybe we can use this PR to just do small tweaks (spacing, colors, etc) and do the animation stuff later on? We could also explore some libs as Hunter was saying (only if we plan to use more animations). If it's only going to be a couple of them, maybe we can do them manually (as you were doing it here, but sometimes system/browser compatibility is a pain :D)

Also interested on what @ChangoMan thinks about all this.

andrealbiac:
Okay, yeah it was to close it up before I open the other issues, but sure we can keep it for future tweaks.

> I'm the guy with the scroll issue haha

When you said this I assumed none of scroll stuff for you (no matter you work on compatibility) so I just ignored it, sorry 😅 
But I'll leave it for now then to focus on the other changes!

andrealbiac:
Okay, thanks for the context!

Then I'd say I revert the parallax changes, keeping only the spacing tweaks and color (I think they include 1 or 2 SVG changes but they're color tweaks). If you think the sliding tab animation is better to have with a library so it's cleaner I can revert that too

and then we can close it up, cause the next PR on this page will be for bigger changes like adding a new section, so it makes more sense to have those in another PR 👍🏻

carletex:
Sounds great!

> If you think the sliding tab animation is better to have with a library so it's cleaner I can revert that too

Up to you!

Thanks Andrea.

PR #211, opened 2025-06-26
Allow to use /builders/ens. Redirect to /builders/address

### LINKED ISSUE #187: Allow to use Ens for builder page:`/builders/{ens}`, not only `/builders/{address}`
For cases when Ens is cached.

I think it should look like `/builders/0xrinat.eth`, but not `/builders/0xrinat`

Both `/builders/{ens}` and `/builders/{address}` should show the same page. I think there should be no redirects address -> ens or ens -> address.

What do you think?

rin-st:
Allow to use `/builder/{ens}`. Redirect to `/builder/{address}`

I agree with Carlos that redirect from ens to address is better https://github.com/BuidlGuidl/SpeedRunEthereum-v2/issues/187#issuecomment-2955083262. And also in case we want to redirect address -> ens when adding ENS to user, previous address page could be cached, so we will have two similar pages, /builders/address and /builders/ens.

Currently, robots.txt disallows /builders/ which means search engines won't index these pages anyway. But I updated metadata in case we want to allow /builders/ in robots.txt in the future

Also, 
- changed parameter /builders/[address] -> /builders/[addressOrEns]
- changed not found description for builders
- changed getEnsOrAddress function and renamed it, so now it uses cached ENS. 
- added one TODO for update ens avatars logic later

To test:
- update your db so users have cached ENS
- check /builders/[addressOrEns] with regular address
- check with cached ens. Should redirect to address
- check with not cached ens. Should show user not found
- check with random text like /builders/asdf. Should show user not found 

Fixes #187


carletex (commented):
Thanks for this Rinat.

My initial reaction when I see a "big" PR (I know, most of it are naming changes, moving files, etc) is to get defensive haha. Especially for a change that it's not a priority.

But after reviewing a couple of times, my defenses are down :P

> getEnsOrAddress => getShortAddressAndEns.

Yeah, this is good!

> builders/[addressOrEns]

This is where I still have my doubts. It makes sense, but it adds some weird abstraction where it also mixes the component with routing logic.

Redirects of this kind are the type of thing that goes into a middleware. I've never use them in NextJS, so not sure how they are. Could we explore them really quick?

rin-st:
> Redirects of this kind are the type of thing that goes into a middleware. I've never use them in NextJS, so not sure how they are. Could we explore them really quick?

Makes sense, middleware feels better for this case.

I had to use api call since `getUserByEns` under the hood uses `node:crypto` which doesnt work with edge runtime. Middlewares work on edge runtime by default. We have nextjs 14, so we cant switch runtime https://nextjs.org/docs/14/app/api-reference/file-conventions/middleware#runtime. It’s possible in nextjs 15, but 
it experimenal though https://nextjs.org/docs/14/app/api-reference/file-conventions/middleware#runtime

> builders/[addressOrEns]

I renamed it back since now it always address when page loads it's always address

packages/nextjs/app/api/users/by-ens/[ens]/route.ts
Comment (carletex) [inline packages/nextjs/app/api/users/by-ens/[ens]/route.ts:1]: Just a random thought (don't need to implement, just saying what is in my mind):

Maybe we shouldn't get the cached value here, and always get the fresh data from ENS. We are kind of resolving ENS to an address.

In the future, it could even be not under /builder, but just `speedrunethereum.com/carletex.eth` and it redirects to `/builders/[address]`.

A lot of PROS / CONS to discuss here, just wanted to mention it.

What do you think, Rinat?

carletex (approved):
Good stuff!!!

Left a comment to consider for the future.

PR #216, opened 2025-07-01
Fix scroll loading on all builds page

ChangoMan:
Replace the old scroll detection with `IntersectionObserver` which is more modern and may be more reliant for scroll detection.


carletex (approved):
Thanks @ChangoMan 

working great!

PR #220, opened 2025-07-04
Create fetchOnchainData service + use waitUntil in user register route

### LINKED ISSUE #218: Register route: wrap ENS fetching logic in waitUntil + create service
Currently, we are fetching ENS data from a user when they register. We are waiting for those request to complete before registering the user and returning.

This is ok now (I might add just < 1 s delay), but thinking in the **future, let's do two things.

1. Wrap it into a `waitUntil` so we can register the user and return early
2. Extract that logic into a service / util

This is the first iteration of this.... but:

## Future

In the new user profiles that we are thinking of, there will be **side quests** where we'd need to fetch onchain data for the user, e.g:

- Have they deployed a contract?
- Have they swapped tokens using DEX?
- Have they held stablecoins?
- Have they used an L2?
- Have they sent a mainnet tx?

We'd need to cache this (we don't to get fresh data every time someone visits a profile). We can use alchemy / zerion API for some of this.

Raw workflow that I have in mind:

1. User register
2. We trigger (inside `waitUntil`) this "fetchOnchainActivity" service where we fetch all the needed stuff (ENS, tx info, etc)
3. We save this in the database (schema TBD)
4. We also have a `lastFetchOnchainActivity`. Initially, it would be `null` (when they register, until the fetch is completed) => so we can show a nice loading the first time they register and visit their profile. Then when the service is done, we just timestamp it.
5. We allow people to refresh this data (♻ icon) ... and we could also have a cron if we wanted.

any thoughts @technophile-04 @Pabl0cks @rin-st ?

carletex:
Fixes #218 

1. Use `waitUntil` to fetch ENS data after user registers
2. Created an onchainData service (more context in https://github.com/BuidlGuidl/SpeedRunEthereum-v2/issues/218). It might be early optimization / a bit verbose... but I think it'll play nicely when we start adding more onchain fetchers services.

### DIFF
--- a/packages/nextjs/app/api/users/register/route.ts
+++ b/packages/nextjs/app/api/users/register/route.ts
@@ -39,11 +39,11 @@ export async function POST(req: Request) {
     const user = await createUser(userToCreate);
 
     // Background processing
+    waitUntil(trackPlausibleEvent(PlausibleEvent.SIGNUP_SRE, {}, req));
+
     waitUntil(
       (async () => {
         try {
-          await trackPlausibleEvent(PlausibleEvent.SIGNUP_SRE, {}, req);
-
           const { ensData } = await fetchOnchainData(address);
 
           // Update user with ENS data if we have any
--- a/packages/nextjs/app/api/users/register/route.ts
+++ b/packages/nextjs/app/api/users/register/route.ts
@@ -4,8 +4,8 @@ import { InferInsertModel } from "drizzle-orm";
 import { users } from "~~/services/database/config/schema";
 import { createUser, isUserRegistered, updateUser } from "~~/services/database/repositories/users";
 import { isValidEIP712UserRegisterSignature } from "~~/services/eip712/register";
+import { fetchOnchainData } from "~~/services/onchainData";
 import { PlausibleEvent, trackPlausibleEvent } from "~~/services/plausible";
-import { publicClient } from "~~/utils/short-address-and-ens";
 
 type RegisterPayload = {
   address: string;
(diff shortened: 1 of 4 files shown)


packages/nextjs/services/onchainData/ens.ts
```diff
@@ -0,0 +1,27 @@
+import { publicClient } from "~~/utils/short-address-and-ens";
+
+export type EnsData = {
+  name: string | null;
+  avatar: string | null;
+};
+
+export async function fetchEnsData(address: string): Promise<EnsData> {
```
carletex, line 8: naming: we are still using `get` and `fetch`. Not sure if we agree in something, but I think `fetch` makes sense when we are making an external request (and get for database calls, etc)

packages/nextjs/services/database/repositories/users.ts
```diff
@@ -234,6 +234,7 @@ export async function updateUser(
     batchId?: number;
     batchStatus?: BatchUserStatus;
     ens?: string;
+    ensAvatar?: string;
```
carletex, line 237: Should we use a TS Partial (+ Omit if needed) here?

packages/nextjs/app/api/users/register/route.ts
```diff
@@ -36,31 +36,33 @@ export async function POST(req: Request) {
...
-      try {
-        const ensAvatar = await publicClient.getEnsAvatar({ name: ensName });
-        if (ensAvatar) {
-          userToCreate.ensAvatar = ensAvatar;
-        }
-      } catch (error) {
-        console.error(`Error getting ENS avatar for user ${address}:`, error);
-      }
-    }
+          const { ensData } = await fetchOnchainData(address);
```
technophile-04, line 48: Shall we split this into two `waitUntil` one for analytics and one for fetching onchain data? 

Because currently if the `trackPlausibleEvent` fails it will not run the `fetchOnchainData`

packages/nextjs/services/database/repositories/users.ts
technophile-04: Ohh yess we could `Partial<UserInser>`?

packages/nextjs/services/onchainData/ens.ts
technophile-04: Makes sense! And we are kind of also following it 🙌 maybe we mentioned this in one of the meet

technophile-04 (commented):
This is working great!! Tysm Carlos!! Just added a small comment regarding splitting of `waitUntil`

packages/nextjs/services/database/repositories/users.ts
carletex: > Ohh yes, the only concern I have is in case in the future, if by mistake we allow an update API endpoint which doesn't have auth, then someone could exploit this function badly.

To be clear, what I was suggesting:

From the user Repository, remove `updateUserSocials`, `updateUserLocation`, `updateUserRoleToBuilder` and leave only `updateUser`

I don't think there is a risk on the endpoints, because the function will only update the fields that are allowed in the function.

e.g. in join-batch endpoint we are doing

```ts
    const updatedUser = await updateUser(user.userAddress, {
      batchId: latestOpenBatch.id,
      batchStatus: BatchUserStatus.CANDIDATE,
    });
```

So it uses the generic repo function but the fields are fixed (no way to hack the update function)

But in update-location, we are doing:

`const updatedUser = await updateUserLocation(userAddress, location);`

But we could use the updateUser function directly.

Make sense?

PR #221, opened 2025-07-08
Track referrer

### LINKED ISSUE #209: Persist visitor referer & UTM data to track delayed conversions correctly
After noticing very low challengeSubmission conversions on our referral sources (ethereum.org, blockscout), and most of our conversion attributed to "Direct / None" source, did some research and testing today to dig into this weird data.

### Problem

Plausible’s short session window (30 minutes) and privacy-first model, resets visitor identity daily (a hash of browser header and ip). This means:

* They reset users every 24 hours for unique visitor stats. Not a big deal
* The referer and UTM attributes live in the user session (lost after 30 minutes inactivity), This means that if a user clicks a link but converts after 30+ minutes (or returns days later), their conversion is attributed to "Direct / None".

As a result, we have no reliable attribution for delayed conversions, which makes marketing ROI analysis a impossible quest.

### Rough solution: Persist registration-time attribution and reuse it for later events

- When a visitor clicks **Register** we capture `document.referrer` and any present UTM parameters.  
- We store those values in the `users` table (e.g. `first_referrer` and maybe a json data field with: `utm_source`, `utm_medium`, `utm_campaign`, `utm_content`, `utm_term`).  
- Whenever the same user submits a challenge, the backend reads these fields and adds them to the Plausible event payload.  

What do you think? Maybe this is not the right solution, but would be great to find a way to better track challengeSubmission conversion origin.

carletex:
As #209 says, the original referrer information was being lost when users returned to the site after a period of inactivity, as Plausible only tracks the referrer at the start of a session (lost after 30 min). This resulted in new sessions being marked as "Direct" unless the user came from another site again. 

**This is a first iteration / starting point (I haven't tested properly) @Pabl0cks will check later and we can keep discussing tomorrow.**

This PR tries to solve the issue:
- store the referrer in local storage when the user visits from another place
- store it in database (when registering)
- use the referrer when sending a request to PLausible (in backend and frontend). https://plausible.io/docs/events-api. 

Also tagging @technophile-04 so he can double check I'm not doing anything weird

Fixes #209

### DIFF
--- a/packages/nextjs/app/start/_components/StartChallengesButton.tsx
+++ b/packages/nextjs/app/start/_components/StartChallengesButton.tsx
@@ -3,16 +3,13 @@
 import { useCallback } from "react";
 import Link from "next/link";
 import { usePlausible } from "next-plausible";
-import { useLocalStorage } from "usehooks-ts";
 
 export const StartChallengesButton = () => {
   const plausible = usePlausible();
-  const [storedReferrer] = useLocalStorage<string | null>("originalReferrer", null);
 
-  // TODO: test this later
   const handleCtaClick = useCallback(() => {
-    plausible("cta_from_start", { props: { referrer: storedReferrer || "" } });
-  }, [plausible, storedReferrer]);
+    plausible("cta_from_start");
+  }, [plausible]);
 
   return (
     <Link
--- a/packages/nextjs/app/start/_components/StartChallengesButton.tsx
+++ b/packages/nextjs/app/start/_components/StartChallengesButton.tsx
@@ -3,14 +3,16 @@
 import { useCallback } from "react";
 import Link from "next/link";
 import { usePlausible } from "next-plausible";
+import { useLocalStorage } from "usehooks-ts";
 
 export const StartChallengesButton = () => {
   const plausible = usePlausible();
+  const [storedReferrer] = useLocalStorage<string | null>("originalReferrer", null);
 
   // TODO: test this later
   const handleCtaClick = useCallback(() => {
-    plausible("cta_from_start");
-  }, [plausible]);
+    plausible("cta_from_start", { props: { referrer: storedReferrer || "" } });
+  }, [plausible, storedReferrer]);
 
   return (
     <Link
--- a/packages/nextjs/components/ReferrerTracker.tsx
+++ b/packages/nextjs/components/ReferrerTracker.tsx
@@ -1,0 +1,30 @@
+"use client";
+
+import { useEffect } from "react";
+import { useLocalStorage } from "usehooks-ts";
+
+// Store the original referer in local storage to track the source of the user
+const ReferrerTracker = () => {
+  const [storedReferrer, setStoredReferrer] = useLocalStorage<string | null>("originalReferrer", null);
+
+  useEffect(() => {
+    // Skip if already stored
+    if (storedReferrer) return;
+
+    // Delete: this is for testing purposes only
+    // e.g. http://localhost:3000/?mockReferrer=https://www.example.com
+    const urlParams = new URLSearchParams(window.location.search);
+    const mockReferrer = urlParams.get("mockReferrer");
+
+    const referrer = mockReferrer || document.referrer;
+    const currentDomain = window.location.hostname;
+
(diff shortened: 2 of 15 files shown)


packages/nextjs/app/start/_components/StartChallengesButton.tsx
```diff
@@ -3,14 +3,16 @@
...
 
 export const StartChallengesButton = () => {
   const plausible = usePlausible();
+  const [storedReferrer] = useLocalStorage<string | null>("originalReferrer", null);
 
   // TODO: test this later
   const handleCtaClick = useCallback(() => {
-    plausible("cta_from_start");
-  }, [plausible]);
+    plausible("cta_from_start", { props: { referrer: storedReferrer || "" } });
```
carletex, line 14: I'm putting here the referrer inside the props. Not sure if that would work (on the backed side, it seems that we can pass it as referrer directly, not inside props)

packages/nextjs/components/ReferrerTracker.tsx
```diff
@@ -0,0 +1,30 @@
...
+// Store the original referer in local storage to track the source of the user
+const ReferrerTracker = () => {
+  const [storedReferrer, setStoredReferrer] = useLocalStorage<string | null>("originalReferrer", null);
+
+  useEffect(() => {
+    // Skip if already stored
+    if (storedReferrer) return;
+
+    // Delete: this is for testing purposes only
+    // e.g. http://localhost:3000/?mockReferrer=https://www.example.com
```
carletex, line 15: Delete this. But useful for testing

cc @Pabl0cks @technophile-04

packages/nextjs/services/plausible.ts
```diff
@@ -10,11 +10,13 @@ export async function trackPlausibleEvent(
   eventName: PlausibleEvent,
   props?: Record<string, string | number | boolean>,
   request?: Request,
+  referrer?: string | null,
 ) {
   const payload = {
     domain: "speedrunethereum.com",
     name: eventName,
     url: `https://speedrunethereum.com`,
+    referrer: referrer || "",
```
carletex, line 19: Check https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/221/files#r2193042006

It seems that this is allowed: https://plausible.io/docs/events-api

packages/nextjs/app/start/_components/StartChallengesButton.tsx
technophile-04: Was just researching on this didn't find any good issues or discussion but while chatting with gpt, it kind of made sense 

Plausible doesn't want you to spoof the user’s real referral data. So if you’re running in the browser, they assume document.referrer is honest. And that's why JS client (and hooks like usePlausible) are meant to be “just fire-and-forget” and hence things like `referrer`, `url` etc are auto captured

technophile-04 (approved):
Thanks Carlos! love the solution and the code looks great to mee!! For some reason not able to see the analytics : 

Image 

Maybe Pablo can just double-check if everything works, but it looks great! We need to run migrate on `main` branch of db merging, tried running it on `staging` branch and everything seems to work 🙌

Pabl0cks (approved):
The `originalReferrer` referer logic is working great to me and storing correctly in database, and our custom events are registering the referrer correctly in dummy tests from staging.

In one of my dummy tests I tried changing the referrer in database, to check if plausible would register the new referrer, but it doesn't seem to work, it keeps attaching the events to the previous referrer. Maybe is because plausible already has a valid referrer and the 30 minute inactivity did not happen?

> I will try in 1 hour ago, but even if this test does not work, we could also update PROD and review in there, just in case it's not working properly in staging, since frontend is not actually registering activity either, only the backend events.

The SRE workflow is working nicely, and register is working nicely with and without referrer. I've deleted the mockReferrer for prod 🙏

image

technophile-04 (approved):

packages/nextjs/app/start/_components/StartChallengesButton.tsx
technophile-04: Got it thankss!

PR #222, opened 2025-07-09
New Builder Profile Challenges

ChangoMan:
Initial revamp of the builder profile page. We are starting by replacing the challenges table with an accordion type design. All potential challenges will be shown to the user, along with a clear progression path.


andrealbiac:
Agree with all Carlos's comments from here and the issue, I was gonna say most of the same things now after seeing the deploy haha

--

I'd add making the titles, submission info (specially icon) and accepted tag bigger, or the challenge description smaller - to have a bit more hierarchy
Captura de pantalla 2025-07-09 a las 13 05 19

--

And also have the START button with square corners (not so rounded) to make different from tags
Captura de pantalla 2025-07-09 a las 13 07 42

--

> * Do we need to create an extra field for the "short short description"? (currently, we are showing the normal description.

And yes, I would make a new short description for each, it feels a bit overwhelming for me with the current one


Lastly, is there a way to be added to the Vercel 'team' or something so I don't have to ask permission to see every time? Not sure if it's possible but would be helpful, thanks 🙏

ChangoMan:
> What should we do with the "State channels" challenge? We were disabling it before (since it's now showing on the homepage)

@carletex I should probably filter out the s"tate channels" challenge, as well as the "stablecoins" challenge. They don't seem to be available on the homepage. It also seems like the "deploy to L2" challenge is "coming soon" so I'll reflect that as well and not link to it.

We just need to make sure this changes to reflect any homepage updates in the future.

> Do we need to create an extra field for the "short short description"? (currently, we are showing the normal description.

We could add a new field, but that might be a pain to keep updated. Another option is that I could line clamp the challenge description, so it maxes out at two lines. I can then add a "read more..." or something to expand it.

The line clamp would be easier to manage, but might not look as pretty.

andrealbiac:
Okay, in any case, wether we edit them or add a new field, we can use these or similar - I already had to crop them for the flyer that we printed in Prague :)

Captura de pantalla 2025-07-10 a las 14 04 43

carletex:
> Just checked and this would work for half of them, the other half we'd need to rewrite. Can you give me some context of why is it hard to maintain? Because you'd need to go to the 'main repo' or database to add a new field and edit it I assume?

As Hunter says, it's not a big deal, but it's a good practice (for our future self) to not add this little "maintenance burdens". (especially if the thing to gain is not very valuable) They keep adding up. 

Personally, the current text doesn't bother me. But I suggested the dot because it's an easy workaround, in case we want a shorter desc

> Just checked and this would work for half of them, the other half we'd need to rewrite. 

I think we just need to tweak the Staking app, no?

image
image

**Layer 2 still showing up!** (I think this is the biggest blocker for the merge)

> The button and tag area were intended to be centered height-wise to the challenge, not aligned to the title. This gives it a bit more space (will work better when the description is shorter too), but not sure how hard it is cause I guess it affects to the whole structure of the entry.

This looks good too, in case it's an easy change, Hunter!

packages/nextjs/app/builders/[address]/_components/ChallengeDetailsStatus.tsx
```diff
@@ -0,0 +1,67 @@
+import Link from "next/link";
+import { ChallengeStatus } from "./ChallengeStatus";
+import type { MappedChallenges } from "./GroupedChallenges";
+import clsx from "clsx";
```
Comment (rin-st) [inline packages/nextjs/app/builders/[address]/_components/ChallengeDetailsStatus.tsx:4]: This could break since it's not our direct dependency, could you remove it and implement all the related logic without that lib?

packages/nextjs/app/builders/[address]/_components/GroupedChallenges.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const userMappedChallenges: MappedChallenges[] = challenges
+    .map((challenge: Challenges[number]) => {
+      const userChallenge = userChallenges.find(uc => uc.challengeId === challenge.id);
+
+      if (!userChallenge) return challenge;
+
+      return {
+        ...challenge,
+        reviewAction: userChallenge.reviewAction ?? null,
+        submittedAt: userChallenge.submittedAt ?? new Date(),
```
Comment (rin-st) [inline packages/nextjs/app/builders/[address]/_components/GroupedChallenges.tsx:42]: looks like there should be something like `null` instead of `new Date()` too

packages/nextjs/app/builders/[address]/_components/GroupedChallenges.tsx
```diff
@@ -0,0 +1,122 @@
...
+        reviewComment: userChallenge.reviewComment ?? null,
+        contractUrl: userChallenge.contractUrl ?? null,
+        frontendUrl: userChallenge.frontendUrl ?? null,
+      };
+    })
+    .sort((a, b) => a.sortOrder - b.sortOrder)
+    .filter(challenge => {
+      return (
+        challenge.id !== ChallengeId.STATE_CHANNELS &&
```
Comment (rin-st) [inline packages/nextjs/app/builders/[address]/_components/GroupedChallenges.tsx:51]: nit: could be moved to separate Set or array too

rin-st (commented):
Thank you, looking good!

Added some small comments

packages/nextjs/app/builders/[address]/_components/ChallengeDetails.tsx
```diff
@@ -0,0 +1,64 @@
...
+            {challenge.sortOrder}
+          </div>
+          <h2 className="m-0 font-medium lg:text-xl">
+            {!isChallengeLocked && !comingSoon && (
+              <Link href={`/challenge/${challenge.id}`} className="hover:underline">
+                {challenge.challengeName}
+              </Link>
+            )}
+            {isChallengeLocked && challenge.challengeName}
+            {comingSoon && `${challenge.challengeName} - Coming Soon`}
```
Comment (rin-st) [inline packages/nextjs/app/builders/[address]/_components/ChallengeDetails.tsx:45]: nit: potentially could lead to error if `isChallengeLocked` is true (for example user didn't complete dependency challenges) and `comingSoon` is true.

image

this should work 

```
            {!isChallengeLocked && (
              <Link href={`/challenge/${challenge.id}`} className="hover:underline">
                {challenge.challengeName}
              </Link>
            )}
            {isChallengeLocked && challenge.challengeName}
            {comingSoon && ` - Coming Soon`}
```

ChangoMan:
Thanks @rin-st ! Those issues should be resolved

carletex:
I think this is good for a first merge (pushed two tiny things) Let's do a second round in another PR.

Some things that are pending / TBD:

- Adding the progress bar on each section (eg. 2 o 3 completed)
- Short short description discussion: leave as it is, add a new file, strip after the first dot, css-line clamp
- Accordion logic: Currently, the first is always open and the second close (no matter the status of the challenge). Let's discuss what' would be best: e.g. close it if all are completed (+ progress bar) OR always closed if it's not your profile..... 

A general thing: Visually, I think it takes a lot of space. It migth be fine now, but it might be too much when we start adding new things. So we should start thinking a more compact alternative: e.g. leave as it's for your profile, but show badges or a reduced version on other's profiles. Or just a general reduced version. Super open, let's discuss it in another PR / issue.

PR #223, opened 2025-07-09
Plausible: referrer as a custom prop

carletex:
It seems that Referrer property is not working  (from the backend)

They have it here: https://plausible.io/docs/events-api

Our hypothesis: since we are forwarding the `User-Agent` and `X-Forwarded-For` in `trackPlausibleEvent`, Plausible is identifying that user with the session initiated from the front-end (with NO referrer). So plausible ignores our hardcoded referrer and just attaches the user to the already existing session.

I'd be nice to integrate properly (also thinking about future UTM stuff) at some point.... but for now, this PR adds a new `originalReferrer` custom prop, so at least we could filter by it (even if we won't see it in the "Top Source" panel).

### DIFF
--- a/packages/nextjs/app/api/challenges/[challengeId]/submit/route.ts
+++ b/packages/nextjs/app/api/challenges/[challengeId]/submit/route.ts
@@ -69,8 +69,12 @@ export async function POST(req: NextRequest, { params }: { params: { challengeId
     waitUntil(
       (async () => {
         try {
-          const referrer = user?.referrer || "";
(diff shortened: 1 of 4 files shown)


Pabl0cks (approved):
Good job @carletex !! Nice workaround to at least be able to register the data in Plausible, LGTM 🙏

In Plausible UI the conversion rate when filtering by these custom properties is not calculated correctly, but i've registered an [issue ](https://github.com/plausible/analytics/issues/5563) for it, and [another one](https://github.com/plausible/analytics/issues/5564) for the referrer not being registered in our previous implementation.

PR #232, opened 2025-07-15
Add progress bar to grouped challenges

ChangoMan:
Adds a new component `GroupedChallengeTitle` which calculates the completed challenges for the group and adds a progress bar:

Screenshot 2025-07-15 at 10 49 21 AM


carletex (approved):
Great stuff, thanks!

Nothing to do here yet, but just mentioning it: We need to figure out what we do with challenges that can't be completed (no autograder). An idea that was floating around is: move them to "builds" section / ideas. So people can take them as reference but update them as builds (vs challenges)

PR #251, opened 2025-07-24
Remove challenges restrictions

### LINKED ISSUE #245: Remove restrictions for challenges
This is something we've been talking about in the past: removing the requirement to do certain challenges. 

Just had a convo with Austin and he feels that not having restrictions could benefit us. We might be losing some advanced people who would love to see / do more advanced challenges, but they can't because they are locked.

I don't see any salient benefit about having the restrictions (well, maybe only for the batches).

What do you guys think?

carletex:
- Challenges won't have dependencies anymore
   - Removed the database column
   - Removed the checks
 - Also removed the check on "disabled". But still keeping that column (it's being used on the API endpoints) because it can be useful
 - Realized that we weren't checking the deps on the backend :D

I know that we could keep some of this "just in case" but I prefer to add it back if needed at some point (also, if there is some specific dependency, we can just do as the batches: hardcoded list using the `getChallengeDependenciesInfo` function)

Let me know if I'm missing something!

Fixes #245

### DIFF
--- a/packages/nextjs/services/database/seed.data.example.ts
+++ b/packages/nextjs/services/database/seed.data.example.ts
@@ -9,7 +9,7 @@ import {
   UserRole,
 } from "./config/types";
 
-export const SEED_DATA_VERSION = "1.1.3";
+export const SEED_DATA_VERSION = "1.1.4";
 
 // Using Drizzle's inferred insert types to ensure seed data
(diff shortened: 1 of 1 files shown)


carletex:
Merging! Let's see if the migration runs automatically as it should (deleting the column)

PR #257, opened 2025-07-26
Update advanced challenges accordion logic

ChangoMan:
Opens up the "Advanced Concepts" accordion if the user has at least 1 challenge completed.


carletex:
Thanks, Hunter!

Let's try with this and see how it feels (definitely better than the current version)

carletex (approved):

PR #258, opened 2025-07-29
Challenge page Header

### LINKED ISSUE #246: Better challenge pages
Right now, it's just the challenge README rendered + Disabled "submit challenge" button

Imagine that we link someone (e.g. from a video) to a specific challenge. I think the page is missing a bit of context of SRE.

Some ideas:

- Iteration v0: Have some tooltip on the "submit challenge" button (connect wallet / register / whatever)
- Iteration v1: Have some kind of header (that only shows if the user is not registered / connected) with some context. example: a text that hints the user to create a profile / submit the challenge / build their portfolio.

carletex:
My v0 attemp to #246 

image
image
image

Fixes #246

### DIFF
--- a/packages/nextjs/app/challenge/[challengeId]/_components/WelcomeBanner.tsx
+++ b/packages/nextjs/app/challenge/[challengeId]/_components/WelcomeBanner.tsx
@@ -1,0 +1,60 @@
+"use client";
+
+import { useAccount } from "wagmi";
+import { RainbowKitCustomConnectButton } from "~~/components/scaffold-eth/RainbowKitCustomConnectButton";
+import { useUser } from "~~/hooks/useUser";
+import { useUserRegister } from "~~/hooks/useUserRegister";
+
+export const WelcomeBanner = () => {
+  const { address: connectedAddress } = useAccount();
+  const { data: user, isLoading: isLoadingUser } = useUser(connectedAddress);
+  const { handleRegister, isRegistering } = useUserRegister();
+
+  if (user || isLoadingUser) {
+    return null;
+  }
+
+  const isWalletConnected = !!connectedAddress;
+
+  return (
+    <div className="w-full max-w-4xl mx-auto mb-8 md:mb-16 p-6 bg-gradient-to-r from-primary/10 to-secondary/10 border border-primary/20 rounded-2xl">
+      <div className="text-center space-y-4">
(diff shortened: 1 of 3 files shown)


packages/nextjs/app/challenge/[challengeId]/_components/WelcomeBanner.tsx
```diff
@@ -0,0 +1,60 @@
...
+            <li>✅ Join our Batch program and connect with other developers</li>
+            <li>✅ Get recognized for your achievements</li>
+          </ul>
+        </div>
+
+        <div className="flex flex-col sm:flex-row gap-3 justify-center items-center">
+          {!isWalletConnected ? (
+            <RainbowKitCustomConnectButton />
+          ) : (
+            <button className="btn btn-primary md:btn-lg" onClick={() => handleRegister(null)} disabled={isRegistering}>
```
Comment (carletex) [inline packages/nextjs/app/challenge/[challengeId]/_components/WelcomeBanner.tsx:44]: Should we redirect to the portolio once the user registers?

I think we don't do it form the existing register button.. but we should?

ChangoMan (approved):
Looks good to me! Tested locally

packages/nextjs/app/challenge/[challengeId]/_components/WelcomeBanner.tsx
Comment (ChangoMan) [reply packages/nextjs/app/challenge/[challengeId]/_components/WelcomeBanner.tsx:44]: It might make sense to redirect on the homepage, but maybe not on a challenge page? Would be an extra step to take them to portfolio if they are already looking and ready to start a challenge.

rin-st:
Thanks Carlos, looking great!

Some nitpicks:
1. "Register as a builder" button is larger than "Connect wallet". Is it intentional?
2. When I'm connected but not registered, I have two invites to register. Maybe remove the right one for the challenges pages if possible?
Screenshot 2025-07-30 at 11 07 16

packages/nextjs/app/challenge/[challengeId]/_components/WelcomeBanner.tsx
Comment (carletex) [reply packages/nextjs/app/challenge/[challengeId]/_components/WelcomeBanner.tsx:44]: Fair! We can leave it as it is right now.

The idea of improving the builder portfolio was to be used as the main CTA (instead of "go to challenge 0" / "submit a challenge"....  It'd be "register")

You will register, get redirected to your profile, and see your new nice profile (with your ENS / onchain data /, sidequest imported + the list of challenges). I think we want to show that "gamification" as soon as possible.

In general, I think it might be best to show/redirect the portfolio, but I kind of agree that maybe not in this case. We can rethink later.

Issue #229, opened 2025-07-14
Remove number from challenge names


rin-st:
We have some challenges with numbers in their names, e.g. Challenge 0 - simple nft example, Challenge 1, 2, 3 etc, and some challenges without numbers - Prediction market, Over collateralized lending, Stablecoins. For some challenges their current number in branch doesn't correspond to their order on sre website - multisig and svg-nfg (I removed challenge number from readmes for now). So I think it's better to remove numbers from challenge names at all, so challenges will not be bound to number and hence we can easily swap order 

Required changes

This repo (To discuss): We're using `sortOrder` to show number of the challenge on main page and profile. Need to think if we need to keep or remove them. For me looks like numbers are not necessary on main page since in future we won't show every challenge there. On the builder profile it helps to understand quantity of the challenges, order, and hence difficulty. But still, we can use just use progress bar and probably it will be enough

se-2-challenges, remove numbers: 
- from readmes
- from branch names
- from test files
- from thumbnails file

grader: 
- change api so it should receive challenge name instead of challenge id (number)
- update `challenges.json` (branch names)
- reinstall challenges

create-eth:
- update `extensions.json` (challenges branch names)

other:
- challenges telegram channels names

cc @carletex @Pabl0cks

rin-st:
I'm thinking to keep this for now

> So, remove numbers
> from main page (?)
> from builder page (?)

and implement all others. What do you think @carletex , can I start to work on it?

rin-st:
[0] done, reverted https://github.com/scaffold-eth/se-2-challenges/pull/334 for `challenge-0-simple-nft` but added change to `challenge-simple-nft`. Added comments to change contract to https://github.com/austintgriffith/speedrun-grader/pull/24, also added new contract to notion

carletex:
Thanks Rinat. This will be easier in person tomorrow, but let me add some comments.

> which we could consider separately.

I feel that if we do it separately, we'll break things because there are dependencies between them. Let me try to propose a more linear approach from items for the different lists.

## 0. Merge the "new" branches

- [ ] https://github.com/scaffold-eth/se-2-challenges/pull/323 - https://github.com/scaffold-eth/se-2-challenges/pull/329. 

**I think we can safely do this right now since it's not being used anywhere**

We could include this here too:
 
- [ ] After merging, we must update https://github.com/austintgriffith/speedrun-grader/pull/24/files#r2225615213 .

## 1. Release the autograder

[...]

PR #259, opened 2025-07-30
Remove Challenge Numbers from the Builder Profile Accordions

ChangoMan:
Remove the challenge numbers `challenge.sortOrder` from the Builder Profile page.

Screenshot 2025-07-30 at 7 09 10 PM


carletex:
Hey Hunter!

What we talked about on https://github.com/BuidlGuidl/SpeedRunEthereum-v2/issues/229 was that we would keep the (display) numbers for now on the site. What we wanted to do is to remove the hardcoded numbers from images, README, branches, etc.... but I think they still contribute in a positive way to the experience (and we can tweak the order in the future without having to change anything on the challenge itseft)

PR #260, opened 2025-07-30
Tweak Start button on homepage

andrealbiac:
Tweaked this button so it looks more like a button and not as much as part of the hero titles, as @phipsae  suggested.

Captura de pantalla 2025-07-30 a las 13 14 44

This is how it looked:
Captura de pantalla 2025-07-30 a las 13 25 23

I can also make it more obvious with square-rounded corners if you think it's still not there. What do you think?


carletex (approved):
Looks great!

just pushed a tweak to remove the new line inside the classes.

thanks

PR #263, opened 2025-07-31
Global error page

### LINKED ISSUE #99: Create global-error page
[Docs](https://nextjs.org/docs/14/app/api-reference/file-conventions/error#global-errorjs)

We need this to avoid crashing the app on server errors and when there is an uncought error on client. Possible [example](https://github.com/BuidlGuidl/SpeedRunEthereum-v2/blob/5307419757de47cea086a8f983656a3307092c3b/packages/nextjs/services/api/users/index.ts#L13) of client error

I think the page should be similar to https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/94

rin-st:
If we have unhandled error on server or client, we have unhandled error page. Something like this:
image

So if you try to add `throw new Error("test");` to any of these places:
- `getAllChallenges()` function
- `page.tsx` - `Home` component
- `HomepageClient.tsx`
and try to load main page, you'll get error page.

`global-error.tsx` fixes it, so when getting error we'll have an our custom error page 
image

Docs: 
https://nextjs.org/docs/14/app/building-your-application/routing/error-handling#handling-errors-in-root-layouts

Also added `error.tsx` since it's recommended
image

and see also
image

Fixes #99

### DIFF
--- a/packages/nextjs/app/error.tsx
+++ b/packages/nextjs/app/error.tsx
@@ -1,0 +1,20 @@
+"use client";
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export default function Error({ error, reset }: { error: Error & { digest?: string }; reset: () => void }) {
+  return (
+    <div className="relative bg-base-300">
+      <div className="absolute inset-0 bg-[url('/assets/home_header_clouds.svg')] bg-top bg-repeat-x bg-[length:auto_200px] sm:bg-[length:auto_300px]" />
+      <div className="relative container mx-auto px-5 mb-11 flex flex-col items-center">
+        <h1 className="mt-16 text-2xl font-semibold md:mt-32 md:text-5xl">Something went wrong!</h1>
+        <button className="btn btn-primary btn-lg md:mt-8" onClick={() => reset()}>
+          Try again
+        </button>
+      </div>
+      <div className="relative h-[130px]">
+        <div className="absolute inset-0 bg-[url('/assets/header_platform.svg')] bg-repeat-x bg-[length:auto_130px] z-10" />
+        <div className="bg-base-200 absolute inset-0 top-auto w-full h-5" />
+      </div>
+    </div>
+  );
+}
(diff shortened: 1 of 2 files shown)


carletex:
Thanks, Rinat!

A couple of questions:
- How does it look like on a live site (I'm guessing that white modal won't show... since those are only for dev)
- Can we show some error trace in the message? It's really useful if someone sends us a screenshot.

rin-st:
Created test PR to check how errors looks on the deployed app, you can check it here https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/264

> How does it look like on a live site (I'm guessing that white modal won't show... since those are only for dev)

Client side error
https://speedrunethereum-v2-git-test-with-error-buidlguidldao.vercel.app/

Server side error
https://speedrunethereum-v2-git-test-with-error-buidlguidldao.vercel.app/builders/0x45334F41aAA464528CD5bc0F582acadC49Eb0Cd1

Error is shown in devtools for both client and server error

---

Server error with error page on builds/[id] page https://speedrunethereum-v2-git-test-with-error-buidlguidldao.vercel.app/builds/1

See builds error.tsx and repositories/db error https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/264/files

Error logs are hidden except [digest](https://nextjs.org/docs/14/app/api-reference/file-conventions/error#errordigest), so we can check vercel log using it.

when tried to log error
image

---

Client error with error page: https://speedrunethereum-v2-git-test-with-error-buidlguidldao.vercel.app/blockexplorer

See blockexplorer page and error.tsx https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/264/files

Error logs are not so helpful. We can show error message on error and suggest to check logs in devtools
image

---

To sum up, I think we need

Check if digest exists
  yes -> server error, just show error.digest
  no -> show error.message and suggest to check devtools logs

This logic is implemented in #264 and if you agree with it, I'll move it to error pages of this PR

rin-st:
> Everything sounds great, let's do it!!

Updated!

carletex (approved):

PR #266, opened 2025-08-01
Add utm parameters tracking to plausible and database

Pabl0cks:
We are currently testing different paid marketing campaigns and need additional tracking data to better understand the ROI of each advertising source. Referrer tracking was a good first step, but we now need to distinguish between paid and organic traffic within the referrer data.

Trying to follow a similar approach than originalReferrer tracking, I've added UTM parameters tracking, so we can have a more granular data on plausible events (register, challenge submission and batch joins).

- We store all UTM parameters in a `jsonb` field that let us query easily in case we want to query statistics directly to database:
  ```
  SELECT * FROM users WHERE original_utm_params->>'utm_source' = 'pabloSource2';
  ```
- When we send plausible events (`joinBatch`, `signupSRE`, `challengeSubmission`) we send also the UTM fields as custom properties toPlausible.

Tested and looks like is registering correctly the new custom properties I created (e.g. `originalUtmSource`) with the stored value, or (none) if user has no original UTM properties stored.

### DIFF
--- a/packages/nextjs/app/api/batches/public/route.ts
+++ b/packages/nextjs/app/api/batches/public/route.ts
@@ -1,0 +1,16 @@
+import { Batch, getAllBatchesDataWithCounts } from "~~/services/database/repositories/batches";
+
+export async function GET() {
+  try {
+    const allBatchData = await getAllBatchesDataWithCounts();
+    const batches = allBatchData.map((batch: Batch) => ({
+      ...batch,
+      telegramLink: undefined,
+    }));
+
+    return Response.json(batches);
+  } catch (error) {
+    console.error("Error fetching sorted batches:", error);
+    return Response.json({ error: "Internal Server Error" }, { status: 500 });
+  }
+}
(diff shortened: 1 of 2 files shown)


packages/nextjs/app/layout.tsx
```diff
@@ -26,6 +27,7 @@ const ScaffoldEthApp = ({ children }: { children: React.ReactNode }) => {
       <body>
         <ThemeProvider enableSystem>
           <ReferrerTracker />
+          <UTMTracker />
```
carletex, line 30: I think we should merge ReferrerTracker & UTMTracker into one! Since they are about the same kind of thing.

carletex (commented):
Looking good, Pablo!

Left a review comment.

Could you also provide a couple of nice URL queries so I can test properly?

Thanks!!

packages/nextjs/app/layout.tsx
Pabl0cks: Merged in https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/266/commits/d655223e170d7fd9eaa8f5fb934852ba5cbbbd85 🙏

carletex (approved):
Thanks, Pablo!!

Tested it locally, and it looks good. I also see your data on Plausible.

Issue #160, opened 2025-05-06
Gate certain api's with API Keys


technophile-04:
### Description: 

Carlos suggested that after #159 it would be hard to get the admin data from scripts. 

So it will nice to allow access to those api's through API keys instead of authorization when quering via scripts

carletex:
Closing this for now until we have the need! (I'm reviewing open issues)

thanks!

Pabl0cks:
I was checking how to get fresh data for [buidlguidl.com/batches](https://buidlguidl.com/batches), since it’s still fetching data from BG3.5, and I realized that `https://speedrunethereum.com/api/batches/sorted?start=0&size=10&sorting=%5B%7B%22id%22%3A%22createdAt%22%2C%22desc%22%3Atrue%7D%5D&filter=` is restricted to admins.

@carletex do you think we should implement API keys for this, or create a new endpoint with the same data as the BG3.5 one just for this integration?

Sample output from the old endpoint:

```
{"id":"3NMQUrEuDBJzyvZZrNc2","name":"7","status":"closed","startDate":1720435180000,"telegramLink":"https://t.me/+oueOa47o0owyOTVh","contractAddress":"0x216E6213DE1740a3f02dd24B1A08DCe95C82dE84","totalParticipants":7,"graduates":7}
```

cc @phipsae

PR #271, opened 2025-08-05
Batches public endpoint

carletex:
We want an accessible endpoint for Batch data (see https://github.com/BuidlGuidl/buidlguidl.com/issues/73)

More context: https://github.com/BuidlGuidl/SpeedRunEthereum-v2/issues/160#issuecomment-3152771662

We could also consider gating this kind of thing under an API key (but let's start with this for now)


Pabl0cks (approved):
Thanks Carlos! Looks good to me 🙌

With this endpoint we should be able to tackle https://github.com/BuidlGuidl/buidlguidl.com/issues/73

cc @portdeveloper

PR #272, opened 2025-08-06
Fixes blockquotes

rin-st:
Removes quotes and italic font-style 

See how it was before in the [issue](https://github.com/scaffold-eth/se-2-challenges/issues/307)

After:
image

Fixes https://github.com/scaffold-eth/se-2-challenges/issues/307


rin-st:
updated it, thanks!

image

Issue #247, opened 2025-07-21
User start workflow


rin-st:
Not sure if it was discussed or if it has a proper issue, but I didn't find it

For me it's not clear one thing:

User goes to sre website. He sees start page by default , right? Which will be `speedrunethereum.com/` He completes some challenges/quests etc. He closes website tab. Tomorrow he returns to the website (open in new tab). What will he see, start page again? We will need "Profile" button on that page then. Or we redirect him to profile if he is registered or completed any challenge/quest? If we redirect him, how he can go to `/start` page again?

---- 

Possible alternative way how it could work

Default page is user profile. If user is not connected or has not completed anything we redirect to `/start` page. 

What do you think?

rin-st:
In all the cases, when opening speedrunethereum.com user will see the page he needs only on the first visit. Well, it's not a big deal, just requires a few additional seconds to click the button.

>  I think we are already doing that with the header menu (that shows "portfolio").

Yes, this is the best and probably is enough. Just thinking that the user will search for the button on the landing first


As an option, I think we can check if challenge 0 is completed and change this text and link to portfolio, and maybe make the background and/or top color of the block more attractive
<img width="410" height="271" alt="Image" src="https://github.com/user-attachments/assets/79c31a36-9b00-4382-af28-97658c9f973f" />

carletex:
> As an option, I think we can check if challenge 0 is completed and change this text and link to portfolio, and maybe make the background and/or top color of the block more attractive

like!

andrealbiac:
I think this is a great point and I would also make the portfolio page the main one for users who already have started SRE. And then you go back to Homepage by clicking on the logo on the header. Not sure how complicated that is to make but it also makes sense to me.



> An alternative will be to show a bigger "header" section with some data about you / your profile (like challenges completed, link to profile, etc)

This is a nice alternative too

rin-st:
Closing this for now

PR #275, opened 2025-08-07
ProgressBasedPrompt component

rin-st:
Changes text for users who already started to do some challenge

image

mentioned [here](https://github.com/BuidlGuidl/SpeedRunEthereum-v2/issues/247#issuecomment-3101808139)

Regarding window color, I think we need to do it in other PR. It requires to move svg to component and add possibility to change colors. Also, I'm thinking that colors of this and other similar svgs on the page should change depending on dark/light theme. What do you think?


andrealbiac:
> Agree! Maybe after #268

Same, I had the same thought

PR #282, opened 2025-08-11
Fix portfolio breakpoint

rin-st:
Micro PR which fixes the grid breakpoint

Before
Screenshot 2025-08-11 at 12 57 02

After
Screenshot 2025-08-11 at 12 55 32


carletex (approved):
Looks good, thanks Rinat!

PR #290, opened 2025-08-13
wrap trackPlausibleEvent to try catch

### LINKED ISSUE #278: Autograded failed after trackPlausibleEvent timed out
I was testing a PR locally, submitted a challenge and got:

> There was an error while grading your submission. Please try again later.

The reason is that `trackPlausibleEvent` timed out.

## Expected behaviour

We should ignore the errors from `trackPlausibleEvent`. It should affect any part of the applications. 

Let's explore what is the best way.

rin-st:
not logging error again since it's already logged inside `trackPlausibleEvent`

fixes #278


packages/nextjs/app/api/challenges/[challengeId]/submit/route.ts
```diff
@@ -72,19 +72,27 @@ export async function POST(req: NextRequest, props: { params: Promise<{ challeng
...
+                originalUtmContent: originalUtmParams?.utm_content,
+              },
+              req,
+            );
+            // eslint-disable-next-line @typescript-eslint/no-unused-vars
+          } catch (e) {
+            console.error(
+              `Error tracking plausible event ${PlausibleEvent.CHALLENGE_SUBMISSION} for user ${userAddress}`,
+            );
+          }
```
Comment (technophile-04) [inline packages/nextjs/app/api/challenges/[challengeId]/submit/route.ts:95]: I think it's better to have a separate `waitUntil` for `trackPlausibleEvent` and also place that `waitUntil` block after the autograder submission `waitUntil`  so that we first enque the autograder request and then enque the analytics

Just tried it out and seems to work as expected with separte waitUntils 🙌

technophile-04 (commented):
Thanks @rin-st, it already seems to work nicely! Just added a small comment regarding separating out `waitUntil` in challenge submit. 

Also, it makes sense to have logging in catch block if it fails 🙌 

Was researching more about it and found that since we are on next 15 now, it is recommended to use [`after()`](https://nextjs.org/docs/app/api-reference/functions/after)

Image 

maybe just to be safe we can make the `waitUntil` => `after` switch in new PR

packages/nextjs/app/api/challenges/[challengeId]/submit/route.ts
Comment (rin-st) [reply packages/nextjs/app/api/challenges/[challengeId]/submit/route.ts]: Thanks, updated! [335e33c](https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/290/commits/335e33cbeaa01337ca7c1f1161131330a3cbdab6)

PR #293, opened 2025-08-14
Loading signature text + Be consistent with loading states

carletex:
This PR aim to do 2 things:

### 1. Better UX on loading states

- Show notification. Loading when waiting for wallet signature. This is feedback I got from @escottalexander, who was connected with Wallet Connect and trying to submit a challenge, and since the text just said "submitting..", he was waiting for the submission to complete (forgetting the signature in his phone). This can also happen when MM is buggy and doesn't open.
- Always show a spinner as loading text (instead of text or spinner + text) to be consistent

image

### 2. Consistency on hooks

There were some hooks (like `useUserRegister`) that weren't returning the `isLoading` properly (because it had a `handle` function that was awaiting the mutation. I refactor these so we directly return the mutation function

-----

### Misc

- Grabbing the UTM params / referer when registering for other places that are not the header/homepage.


carletex:
> nit: Maybe we could wrap this in the new function so we don't need to repeat it everywhere?

I didn't want to refactor that hard in this PR, but went for it, and I think it looks good! Created a hook (instead of a regular function/util, so we don't have to pass some extra params: like the `useSignTypedData` from wagmi)

Wdyt?

rin-st:
> Wdyt?

I like it!

Retested just in case and it works fine! Thank you!

PR #296, opened 2025-08-14
Fix user points

### LINKED ISSUE #294: Wrong XP calculation
When calculating the XP for a user on `getUserPoints` (maybe we could also rename to `getUserXP` in this PR :)) we are considering ALL the submissions for a challenge, and we should only take in consideration the last one. If not, you can get a bunch of XP for the same challenge.

e.g.

I submitted:
- approved
- approved
- rejected

I should get 0 points, but I'm getting 20.

Image

Image

rin-st:
Fixes #294


carletex (approved):
Thanks, Rinat!!

Yes, this is probably the best way to do it. Doing it in the same query might be too much.

Sadly, I won't be able to do this anymore haha

image

PR #297, opened 2025-08-15
Bigger PunkBlockie. Show XP points

carletex:
Little tweaks to the unfurl image after some feedback from Austin:

- Bigger PunkBlockie (to match the size - 200 - of the ENS Avatar)
- Show XP instead of Challenge completed

image

image

I also renamed the `getUserPoints` function to `getUserXP`.

----

Tested it with the oGraph Preview extensions


rin-st (approved):
Lgtm, thanks!

Issue #241, opened 2025-07-18
Side quests tab


andrealbiac:
<img width="227" height="319" alt="Image" src="https://github.com/user-attachments/assets/9d1664be-6848-47c2-ab2d-7ad467aafaee" />

Questions here:

1. What items to include
2. how do we fetch that data
3. where to place it in the profile

---
1. I liked Pablo's ideas with the ENS name, but I think I would keep it simple so you can still get all the XP points; for example, it's nice that you set up your name and picture in ENS but maybe less likely to register a subdomain? (idk)
2. (can't talk much about this)
3. I was thinking maybe we can move the achievements to the top of the page to leave space for the sidequests on the side, and keep it compact (just an idea):

<img width="810" height="697" alt="Image" src="https://github.com/user-attachments/assets/d9b9c3c1-30d9-4aa7-bb87-cc67bff8a8d5" />

First badge shows title on hover in the image

rin-st:
I like how both the side quests and achievements sections look!

Some thoughts
Probably we need side quests page with a small description how to register ens/avatar and then cache it on SRE. So maybe we also need a small button "Start" or similar on side quests section too. Also, not sure if we need subdomain and primary name. To set primary you need 2 or more ens names, right?

carletex:
> Probably we need side quests page with a small description how to register ens/avatar and then cache it on SRE. 

Yes, we could probably create a Guide! (#26 ) for it! And we could link it from each side quest (question mark icon, or just hover)

> Also, not sure if we need subdomain and primary name. To set primary you need 2 or more ens names, right?

Agree! (I think Andrea just used placeholders)

Some ideas from https://github.com/BuidlGuidl/SpeedRunEthereum-v2/issues/218 (besides ENS name + Avatar)

- Have they deployed a contract?
- Have they swapped tokens using DEX?
- Have they held stablecoins?
- Have they used an L2?
- Have they sent a mainnet tx?

We can check which APIs could be good for getting this data. We can start small

Pabl0cks:
> Yes, we could probably create a Guide! (https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/26 ) for it! And we could link it from each side quest (question mark icon, or just hover)

I really like this idea ♥

andrealbiac:
> Agree! (I think Andrea just used placeholders)

Yeah, and I agree I wouldn't make them too complex so it still feels achievable, otherwise might be overwhelming or loose excitement

[...]

PR #298, opened 2025-08-18
SideQuests: initial snapshot to DB on user registration

Pabl0cks:
Implements sidequests as a database snapshot (jsonb field in users table) saved on user registration. 

## Some context about Zerion API:

- The endpoint [https://api.zerion.io/v1/wallets/{address}/transactions](https://developers.zerion.io/reference/listwallettransactions) fits our needs to check v0 sidequests statuses
- Can check the endpoint website to check all available filters, attaching a response example below
- They offer a free tier of 5000 req/day, 120 req/sec
- They have a limit of 100 transaction per request, so I see a couple of options:
   - Call the endpoint with specific filtering for each sidequest, if we receive items => quest completed
   - Fetch transactions with pagination in batches of 100, until all non-ENS quests are found, or no transaction pages left
- Only using the `filter[trash]` filter at the moment, to fetch `only_non_trash` to avoid spam transactions

Attaching [response example](https://github.com/user-attachments/files/21844746/notetaker_response.txt) for wallet `0x60DAf0e5Ff2eC9BF54c8AbF5623f95287Bcd44fB` executed in their [website](https://developers.zerion.io/reference/listwallettransactions)

## Sidequest completion logic

- ENS completion comes from the database `ensName` (not from Zerion transactions). 
- Detection rules are simple: 
   - Mainnet tx = outgoing on `ethereum` chain.
   - L2 = outgoing on known L2 chains (should we simplify the list or fetch it automatically?)
   - DEX trade = `operation_type === "trade"` and a transfer of a fungible token
   - Stablecoins held = inbound transfer of common stables or a symbol containing “USD” (maybe this check may cause false positives?)
   - Deployed a contract = `operation_type === "deploy"`
   - Minted an NFT = `operation_type === "mint"` and a transfer of an nft token

First PR to tackle https://github.com/BuidlGuidl/SpeedRunEthereum-v2/issues/241


carletex:
Hey Pablo, thanks for putting this together. After our conversations, I did a big refactor of the sidequest feature. Let me give some context, so @rin-st can also review.

You version was _simple_ in a sense that it was in a simple file, but it was hiding a lot of complexity and had a lot of coupling.

I decided to do an architectural refactor (without getting too crazy). My main goal was to make it clear, simple, and maintainable (if we add / remove sidequest). Main piece:

- `sideQuests/schema.ts`: The source of truth for the sidequest config (since we are not using the database to save the config, as in challenges). `SIDEQUESTS` will contain all the data that we need (e.g. link to a guide).
-`sideQuests/checks.ts`: It holds the check function for each side quest.
- `sideQuests/index.ts`: Export the main function to be used `fetchSideQuestsSnapshot`. Take an `user` as a param. Changes from your approach:
  - Only saves completed sidequest to the database. I think `pending` state was a bit confusing, and we were adding a lot of unnecessary data. I think this is enough for the UI too.
  - Skip completed side quest (not sure if you had this)
  - Removed the `sidequest` object "prefix" in the new database field (`sidequests: {data}` => `{data}`)
- `sideQuests/types.ts`: just needed types.

- `onchainData/zerion.ts`: wraps Zerion API calls

----

To test:
- Added Zerion API KEY env var to Vercel (remove the base64 codification in the code, and just using the encoded version in the env var, as it was in the docs)
- You can create this page if you don't want to test by registering all the time:

```ts
// app/sidequest/page.tsx
import { getUserByAddress } from "~~/services/database/repositories/users";
import { fetchSideQuestsSnapshot } from "~~/services/sideQuests";

export default async function SideQuestsPage() {
  // TODO: Replace with the user you like that exist in the database.
  const user = await getUserByAddress("0xB4F53bd85c00EF22946d24Ae26BC38Ac64F5E7B1");
  if (!user) {
    throw new Error("User not found");
  }
  const sideQuests = await fetchSideQuestsSnapshot(user);
  return <pre>{JSON.stringify(sideQuests, null, 2)}</pre>;
}
```

rin-st:
Thanks @Pabl0cks for the PR and @carletex for context and refactor, now it's really well structured! Looking great to me!

PR #299, opened 2025-08-21
Fix socials notification

### LINKED ISSUE #295: Notifications behind the overlay on Update Socials
We fixed it in other modals / overlays, but in this one, the notifications still shows behind

Image

rin-st:
Fixes notification similar to some others like `BuildFormModal`

image

Fixes #295


carletex (approved):
Thanks!!

PR #300, opened 2025-08-21
Fix disconnect with walletConnect

### LINKED ISSUE #291: Bug: Can't disconnect wallet on challenge pages
This video shows the problem
[YouTube video](https://youtu.be/zk3qGrCssl0)

I was able to disconnect on the main home page but not on any challenge pages I tried.

Extra info that may be relevant: My wallet was connected with wallet connect.

rin-st:
Bug appeared because of race condition between `disconnect()` and `signOut()`. Changed to `async` versions and bug disappeared.

Tested with WalletConnect and Metamask app on my phone (android). Note: Sometimes signing messages doesnt work if you open MM app **after** you click button, which requires signing. So you need to keep MM app open to successfully sign messages

Also see my comments in PR

Fixes #291 

cc @escottalexander


packages/nextjs/app/admin/_components/AccessDenied.tsx
```diff
@@ -16,9 +16,9 @@ export const AccessDenied = () => {
         </Link>
         <button
           className="btn btn-primary border-2 border-primary"
-          onClick={() => {
-            disconnect();
-            signOut();
+          onClick={async () => {
```
rin-st, line 19: same bug was here, changed to async and now it works as expected

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton/index.tsx
```diff
@@ -25,17 +25,21 @@ export const RainbowKitCustomConnectButton = () => {
...
 
   useEffect(() => {
-    if (sessionUserAddress && connectedAddress && connectedAddress !== sessionUserAddress) {
-      disconnect();
-      signOut({ redirect: true, callbackUrl: "/" });
-    }
-  }, [connectedAddress, sessionUserAddress, disconnect]);
+    const disconnectIfWrongAddress = async () => {
+      if (sessionUserAddress && connectedAddress && connectedAddress !== sessionUserAddress) {
+        await disconnectAsync();
```
rin-st, line 36: couldn't reproduce this with walletConnect, changed just to be safe

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton/WrongNetworkDropdown.tsx
```diff
@@ -20,7 +21,10 @@ export const WrongNetworkDropdown = () => {
           <button
             className="menu-item text-error btn-sm !rounded-xl flex gap-3 py-3"
             type="button"
-            onClick={() => disconnect()}
+            onClick={async () => {
```
rin-st, line 24: couldn't reproduce this with walletConnect, changed just to be safe. + added `signOut()`

packages/nextjs/components/scaffold-eth/RainbowKitCustomConnectButton/index.tsx
```diff
@@ -69,7 +73,10 @@ export const RainbowKitCustomConnectButton = () => {
           return (
             <div className="flex items-baseline gap-4">
               <RegisterUser />
-              <button className="flex items-center rounded-full bg-base-300" onClick={() => disconnect()}>
+              <button
+                className="flex items-center rounded-full bg-base-300"
+                onClick={async () => await disconnectAsync()}
```
rin-st, line 78: changed this just to avoid two imports 

```
const { disconnect, disconnectAsync } = useDisconnect()
```

technophile-04:
For some reason, I wasn't able to reproduce it :(...Like I tried opening SRE on my laptop and connecting with walletConnect from my MM mobile app and it disconnect nicely. 

Using: 
Arc (chrome based browser) 
mac
Iphone 

Also tried opening SRE in MM mobile app it also works expected there
 
also from https://github.com/BuidlGuidl/SpeedRunEthereum-v2/issues/291#issue-3319858794 

> I was able to disconnect on the main home page but not on any challenge pages I tried. 

This is strange behaviour since same component is rendered in every page :( 

--- 

But I think the changes makes sense!! Maybe @rin-st we can also wrap the `async` in try catch and log error so that next time any simillar bug happens, we can get some hints from console?  

Thanks Rinat!

rin-st:
> I was able to disconnect on the main home page but not on any challenge pages I tried.

> This is strange behaviour since same component is rendered in every page :(

For me disconnect from main page also was bugged, sometimes it worked but sometimes is not.

> But I think the changes makes sense!! Maybe @rin-st we can also wrap the async in try catch and log error so that next time any simillar bug happens, we can get some hints from console?

Thanks! Updated

Pabl0cks (approved):
I wasn't able to reproduce either :/

Disconnect was/is working nicely to me from challenge page and homepage with this setup:

- Windows laptop + Chrome
   - WalletConnect with rainbow wallet on mobile (android)
   - WalletConnect with safe wallet from the same laptop

technophile-04 (approved):
Merging this sicne it doesn't have side effects and also if next time something similar happens #291 we can get logs  🙌

PR #305, opened 2025-08-26
Use nextjs after instead of waitUntil

rin-st:
from https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/290#pullrequestreview-3116976824

cc @Pabl0cks just to letting know that plausible tracking code will be changed a bit


technophile-04 (approved):
Looking great! Thanks @rin-st! We are having merge conflict at `register` route...I think we can add `fetchOnChainRequest` and `fetchSideQuest` block in new after blocks

rin-st:
> Looking great! Thanks @rin-st! We are having merge conflict at register route...I think we can add fetchOnChainRequest and fetchSideQuest block in new after blocks

Updated! `fetchOnchainData` only, it's enough [763240b](https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/305/commits/763240b3474afdd7d32484b21441f74ac586665f)

PR #310, opened 2025-08-28
Sidequests ui

rin-st:
Screenshot 2025-08-28 at 16 53 17

To test open portfolio of the user with `sideQuestsSnapshot`

### DIFF
--- a/packages/nextjs/app/builders/[address]/_components/PointsBar.tsx
+++ b/packages/nextjs/app/builders/[address]/_components/PointsBar.tsx
@@ -52,7 +52,7 @@ export function PointsBar({ points, totalPoints }: { points: number; totalPoints
   const modalRef = useRef<HTMLDialogElement>(null);
 
   return (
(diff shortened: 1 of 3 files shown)


carletex:
Great stuff, Rinat!

Sorry for pushing directly, but I tried https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/310/commits/172e82cc5dcd899b9f08f9cb7383faf1757bf0d0. Using string (even if typed) looked a bit weird to me. Not sure if there is a better way (without repeating the `id` in the `SIDEQUEST` object). Does it make sense to you? 

Also, a couple of things I think we should add to this PR:

- ENS Avatar side quest. It should be easy to add.
- Use sidequests for XP points calculation. What about adding 5 points per sidequest completed?

Maybe we should fix and merge #307 first?

carletex:
Merging this, good stuff!!

Next, I think we could add some guides linked to some sidequest, or/and a more detailed description (on hover).

PR #311, opened 2025-08-29
Script to update onchain data for existing users (+ throw errors instead of error silently)

carletex:
(**This is not ready for merge, it's just a way to open the discussion.**)

After #310, sidequests are visible to everybody. I think it'd be nice to pre-populate it for all the existing users. New users will get it by default (on the register route). And yes, existing users might have outdated sidequests when they log in 3 months from now...but they will have the ♻. And in the future, we can have a cron job if we want.

I crafted a script just for testing with 5 users (with 10, I started to get a lot of errors). From Zerion API:  «For development keys, the limit is currently set at 5k requests per day. This is spread across ~60 requests per minute with a 30 seconds waiting period» (I don't think it's worth it to request an upgrade for the key... since this is a one time thing. Unless we implement a cronjob in the future)

The script can be improved with:
- `setInterval`. Process 5/8/10 users (we are making 5 Zerion requests per user) every minute.
- Database batches. Write all the data in one tx (per "interval") to the database. (I think we need to use `executeQueries`)

## On Try/catch, throwing and silent errors

Something to keep in mind: Only _orchestrator/top-callers/top-consumers_** should "eat" the errors. We had a lot of try/catch with `console.error` and returning a valid value (like `false` or `[]`) inside helper functions.

** orchestrator/top-callers/top-consumers: routes, scripts, etc. 

This is a big problem, because if something fails (e.g. a Zerion call), the top-consumer won't be able to know that an error happened (we were just showing an error in the console, which the orchestrator can't detect).

It should always be up to the orchestrator to decide what to do with the error (e.g., return a 400 error, eat the error, etc). So helper functions should always:

1. Not use try/catch at all (let errors bubble up naturally)
2. Use try/catch, log the error and pass the error by throwing again

The only benefit (I think) for 2, is if we want to console.error for context reasons (or protect in case the orchestrator forgets...not sure if this is a good reason). In any case this is not a big deal if we re-throw (the big problem are the silent errors)

I also think that in most codebases, people omit try/catch in these helper functions (and also have a general app-wide try catch in case no one catches the error (I think Rinat did this in our Next app). 

A good example:

```ts
export async function checkContractDeployed({ user }: SideQuestCheckArgs) {
  try {
    const txs = await fetchTransactions(user.userAddress, { "filter[operation_types]": "deploy" });
    return txs.length > 0;
  } catch (err) {
    console.error("contractDeployed check failed", err);
    throw err;
  }
}
```

to

```ts
export async function checkContractDeployed({ user }: SideQuestCheckArgs) {
  const txs = await fetchTransactions(user.userAddress, { "filter[operation_types]": "deploy" });
  return txs.length > 0;
}
```

We don't really need the try/catch. If error, this will bubble up to the corresponding orchestrator, and it will decide what to do.

In this particular case: script/route -> fetchSideQuestsSnapshot -> checkContractDeployed. 

- checkContractDeployed. No point in having a try / catch
- fetchSideQuestsSnapshot. Might make a bit more sense to have try/catch to be more verbose on the error (but not needed) -> re-throw
- script/router: this should definitely have the try / catch... and decide what to do.

Updated with 6414c7e34b5e36308de227b6221648c4d809817e just to see how it feels

Would love to hear what you guys think


rin-st (approved):
Lgtm!

technophile-04 (approved):
LGTM to me as well! Merging this 🙌 TYSM Carlos!

PR #315, opened 2025-09-03
Updated purple in challenge images

andrealbiac:
Quick pr to update the challenges that use purple- making a new one for ZK Voting (that I didn't upload yet bc the challenge is still not ready) uses the purple too and I realised the one we were using was a bit too light against the blue background. So I darkened it slightly:

social card-zk


carletex (approved):
Ok, thanks!

We'll also need to tweak them in each challenge READMEs. https://github.com/scaffold-eth/se-2-challenges

Can you make that happen? if not, you can send the assets and we can replace them.

Thanks!

Note: State channels challenge is not part of the curriculum anymore.

PR #330, opened 2025-09-30
add correct tokenization image

escottalexander:


carletex:
oh, had access to the Figma, so I went for it.

Merging!

PR #334, opened 2025-10-02
Fix font color for How is XP calculated modal item titles

### LINKED ISSUE #333: How is XP calculated modal titles are hard to read in dark mode
The titles are hardcoded to gray and it is not very readable in dark mode.

Image

Pabl0cks:
Fixes #333 

Previous:
image
image

New:
image
image


rin-st (approved):
Lgtm!

PR #339, opened 2025-10-22
Private notes

### LINKED ISSUE #329: Add private notes field for builders (admin-only)
Hi,

I’d like to suggest adding a database field where we can store more detailed information about each builder. A simple string/notes field per builder would already help a lot (maybe like in grants). It would make tracking easier than relying only on Google Sheets. This field should only be visible to admins and not exposed via the public API. Thinking of it like a lightweight CRM.

I’m also considering introducing a new builder role, something along the lines of “scammer/hacked” (naming still to be decided). The idea would be to flag compromised or malicious addresses so we can keep the database cleaner.

What do you think? Open to other suggestions as well. The 2nd point with role is lower priority for me.

Thanks

carletex:
Adds private notes to the system (admins can add private keys to builders)

image
image

Fixes #329

### DIFF
--- a/packages/nextjs/app/builders/[address]/_components/UserNotesIndicator.tsx
+++ b/packages/nextjs/app/builders/[address]/_components/UserNotesIndicator.tsx
@@ -25,7 +25,7 @@ export const UserNotesIndicator = ({ user }: UserNotesIndicatorProps) => {
     <label
       htmlFor={USER_NOTES_MODAL_ID}
       className="absolute top-[-15px] right-[-15px] btn btn-sm btn-circle btn-warning cursor-pointer tooltip tooltip-left items-center justify-center flex"
-      data-tip="private notes"
+      data-tip="Private notes"
     >
       <div className="">{notesCount}</div>
     </label>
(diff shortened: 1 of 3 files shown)


phipsae:
@carletex nice implementation, really cool, thanks a lot! :)

packages/nextjs/app/builders/[address]/_components/UserNotesIndicator.tsx
```diff
@@ -0,0 +1,33 @@
...
+    );
+  }
+
+  const notesCount = notes?.length || 0;
+
+  return (
+    <label
+      htmlFor={USER_NOTES_MODAL_ID}
+      className="absolute top-[-15px] right-[-15px] btn btn-sm btn-circle btn-warning cursor-pointer tooltip tooltip-left items-center justify-center flex"
+      data-tip="private notes"
```
Comment (rin-st) [inline packages/nextjs/app/builders/[address]/_components/UserNotesIndicator.tsx]: Maybe it's better to capitalize first char? `Private notes`

rin-st (approved):
Lgtm, thank you!

Added some small nitpicks

carletex:
Thanks all!

Addressed Rinat's feedback  here https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/339/commits/51ded16f553d5db551f4e74a563ec593973c95fa

PR #342, opened 2025-10-29
Shorten /start page CTA to fix display when wallet not connected

### LINKED ISSUE #341: Fix /start CTA to first challenge when you're not connected
Fix the CTA container to show the link in 1 line

Image

Pabl0cks:
Change `ProgressBasedPrompt` link text to “Jump to the first challenge” so the CTA stays on one line on /start when there is no user connected.

image

Fixes #341


carletex (approved):
Thanks Pablo!

PR #343, opened 2025-11-04
Enhance plausible tracking for registrations

Pabl0cks:
Added tracking to see where users are registering from and which buttons they're clicking to register.

**Changes:**
- Auto-capture the current page path when user registers
- Pass a trigger identifier from each register button
- Send both to Plausible as custom properties
- Send current page as `eventUrl` to show in "Top pages" UI from plausible, currently we were always sending a hardcoded `https://speedrunethereum.com` value

**3 registration points tracked:**
- Header tooltip → `header_register_tooltip_button`
- Start page → `start_register_button`  
- Challenge banners → `challenge_banner_register_button`

The data shows up in Plausible under the signupSRE event with "eventPath" and "eventTrigger" properties you can filter/breakdown by.

### DIFF
--- a/packages/nextjs/app/_components/RegisterUser.tsx
+++ b/packages/nextjs/app/_components/RegisterUser.tsx
@@ -62,7 +62,7 @@ export const RegisterUser = () => {
               handleRegister({
                 referrer: storedReferrer,
                 originalUtmParams: parsedUtmParams,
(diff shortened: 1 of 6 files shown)


carletex (approved):
This looks good, thanks Pablo!

PR #372, opened 2026-03-26
Improve SVG NFT prompt

damianmarti:
Tested with Claude Sonnet 4.6 and Opus 4.6.

Overall, the results from both models are great, without human intervention.

Added some improvements.

 - Switches SVG string assembly recommendation from abi.encodePacked to string.concat (no need to use OpenZeppelin Strings or `abi.encodePacked` anymore).
  - Adds guidance on contract size limits — SVG-heavy contracts are prone to hitting the 24KB limit, with a recommendation to split the renderer into separate contracts/libraries. It was an issue when trying to build a complex SVG art.  Fixed after adding this line.                                                                           
  - Adds EIP-4883 compliance requirement, including the `renderTokenById(uint256 id)` interface that must return valid SVG element(s). This is to allow composability.

Should we add to the prompts to remove YourContract and related stuff?

### DIFF
--- a/packages/nextjs/build-prompts/svg-nft.md
+++ b/packages/nextjs/build-prompts/svg-nft.md
@@ -35,7 +35,11 @@ A smart contract that generates and renders SVG artwork entirely on-chain. Each 
 - **Required Views:** Frontend to mint, view, and transfer NFTs. A personal gallery view to see owned or all minted NFTs.
 - **Agent Autonomy:** Build the `UI_THEME` to let the art shine. The frontend MUST correctly fetch, decode, and parse the base64 data URIs from `tokenURI()` so the SVGs render natively and beautifully in the DOM as images without broken links.
 
-## 6. Next Iterations (Builder: ask the agent to add these later)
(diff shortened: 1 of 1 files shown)


technophile-04:
image


This is great! Thanks @damianmarti! 

Just one thing, we should add "Review the generated code using the grumpy-carlos-code-reviewer agent."  I feel its really helpful

technophile-04 (approved):

Pabl0cks (approved):
Thanks Damu! Looking good to me! ❤️

> Just one thing, we should add "Review the generated code using the grumpy-carlos-code-reviewer agent." I feel its really helpful

Agree, thanks Shiv! I think we can add it to all of the build prompts as an extra section before "Next Iterations": `## 6. Run grumpy-carlos-code-reviewer agent`

damianmarti:
Thanks @technophile-04 and @Pabl0cks for the review.

Added a review step, as the one added on https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/374

PR #373, opened 2026-03-26
Improve commit-reveal prompt

damianmarti:
Tested with Claude Sonnet 4.6 and Opus 4.6.

The contract code looks and works great.

The generated frontend is not so good. 

I think we need to add at least _Show the games from the user and the open games (created games with no opponent)._ to the required views. Adding this, the UI is much better, allowing to find games easily and revealing committed games.

### DIFF
--- a/packages/nextjs/build-prompts/commit-reveal-game.md
+++ b/packages/nextjs/build-prompts/commit-reveal-game.md
@@ -7,26 +7,32 @@ imageUrl: "/assets/build-prompts/commit-reveal-game.png"
 # SPEC: Commit-Reveal Game
 
 ## 1. Objective
+
 A betting game that uses cryptographic hashing to hide a player's move until both have played, preventing blockchain front-running.
 
 ## 2. Pre-flight & Context
+
 - **Framework:** Scaffold-ETH 2.
 - **Check:** Verify if ./packages/nextjs/scaffold.config.ts exists directly in the current working directory (do not search subdirectories). If it doesn't exist, this is not a Scaffold-ETH 2 project. Follow the instructions at https://docs.scaffoldeth.io/SKILL.md to scaffold it first. If it exists, continue directly with building.
 
 ## 3. Builder Customizations (Tweak before prompting)
+
 - `GAME_TYPE`: Rock-Paper-Scissors
 - `UI_THEME`: Arcade style or competitive 1v1 interface.
 
 ## 4. Smart Contract Spec
+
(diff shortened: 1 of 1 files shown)


damianmarti:
Added review step.

packages/nextjs/build-prompts/commit-reveal-game.md
```diff
@@ -7,26 +7,36 @@ imageUrl: "/assets/build-prompts/commit-reveal-game.png"
...
 ## 5. Frontend Spec
-- **Required Views:** A two-step game interface guiding the user securely through Committing and Revealing.
+
+- **Required Views:** A two-step game interface guiding the user securely through Committing and Revealing. Show the games from the user and the open games (created games with no opponent).
 - **Agent Autonomy:** The frontend MUST handle the local generation of the secret salt and the `keccak256` hashing locally before sending the transaction. Hide this complexity from the user for a smooth UX.
 
-## 6. Next Iterations (Builder: ask the agent to add these later)
+## 6. Review
+
+- Always review the code using the grumpy-carlos-code-reviewer agent.
```
technophile-04: nit: Shall we instead say something like: 

- Review the whole generated code using the grumpy-carlos-code-reviewer agent once. 


--- 

Because the current line might trick the agent it's not mandatory to run the review agent. It's just mandatory to use it when asked. 

But I mean when I ran the prompt with opus 4.6 it actually used the reviewer

technophile-04:
Works nicely, Thanks Damu! Just a nit on frontend, I tried running the prompt 2 times and both of the times: 

It had the game details page and all the available games on same page handled by react useState. Maybe we can add an instruction to have a `/game/[id]` page so that the when a person refereshes he doesn't go back to All games page

Pabl0cks (approved):
Nice job Damu, working nicely to me.

The only issue I had is (on an empty folder test) it did not add the bets, only did let me commit and reveal without placing bets on them, but maybe is a gpt5.4 or a codex cli issue, first day trying it.

packages/nextjs/build-prompts/commit-reveal-game.md
damianmarti: Yes, it always worked for me but I understand what you say.

I will change it to:

```
- Always review the generated code and use the grumpy-carlos-code-reviewer agent for that.
```

damianmarti:
@technophile-04 @Pabl0cks thanks for the review!

Added more data to the required views to fix these issues: game page and bet.

Tested after that, and these features were implemented nicely and cleanly.

PR #374, opened 2026-03-30
Improve payment splitter prompt

damianmarti:
Tested with Claude Opus and Sonnet.

Adding an additional step asking the agent to always review the code using the grumpy-carlos-code-reviewer worked great.

The prompt is already very detailed, and it seems enough for the agents to implement this. 

The gaps between what we want and what is implemented are resolved with the review by the grumpy-carlos-code-reviewer agent.

We have to check if we need to add this explicit step to all the prompts, or if we can ensure the review of grumpy-carlos-code-reviewer in some way.


Pabl0cks (approved):
LGTM! 🙌

PR #377, opened 2026-04-08
Update multisig build prompt

rin-st:
- added possibility to revoke transactions
- update se-2 theme in `UI_THEME`
- include `receive()` function in smart contract
- frontend: owner administration panel (add/remove, threshold). Removed tailwind/daisy
- added grumpy-carlos

### DIFF
--- a/packages/nextjs/build-prompts/multisig-wallet.md
+++ b/packages/nextjs/build-prompts/multisig-wallet.md
@@ -8,27 +8,37 @@ featured: true
 # SPEC: Multisig Wallet
 
 ## 1. Objective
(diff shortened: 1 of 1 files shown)


damianmarti (approved):
It worked well, and the code looks good after some fixes from the grumpy-carlos review.


I think we need to add something like "After each transaction is executed, show the result and the link to the transaction detail" to the Required Views item. And after "and a queue displaying pending/executed transactions", add something like "(showing a link to the transaction detail for executed transactions)".

Thanks!

rin-st:
Makes sense! Updated 81e5562 ! Wrote it one time to avoid duplication, and it worked for me

Thanks!

PR #381, opened 2026-04-16
Tweaks to quadratic funding

carletex:
Worked pretty well for me in a few runs.

Just added some tweaks.

### DIFF
--- a/packages/nextjs/build-prompts/quadratic-funding.md
+++ b/packages/nextjs/build-prompts/quadratic-funding.md
@@ -7,17 +7,21 @@ imageUrl: "/assets/build-prompts/quadratic-funding.png"
 # SPEC: Quadratic Funding Pool
 
 ## 1. Objective
-An always-open grant allocation system where a central matching pool is distributed to projects based on the *number* of unique contributors, not just the total amount raised. Anyone can register a project, anyone can donate, and an admin can trigger matching fund distribution at any time.
+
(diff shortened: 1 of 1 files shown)


damianmarti (approved):
Tested with Sonnet 4.6.

Worked great, the code looks good, and the frontend too. All features work as expected.

Thanks!

Pabl0cks (approved):
Looks good to me! (sonnet 4.6, opus 4.7 and gpt5.4 -gemini got out ot tokens haha-)

**Question**
Should we add to the prompt the ability to edit your own projects? You create the project, realize you had some typo or the image is not rendering and want to edit it, right now is not generating this feature. Something like:

A line in Section 4 (Smart Contract Spec), under Project Registration:

```
- Project Editing: updateProject(projectId, name, description, imageURI): only callable by the project's recipient. Allowed while the current cycle is active (i.e., before distributeMatching() is called); reverts otherwise. Emits a ProjectUpdated event.
```

  A line in Section 5 (Frontend Spec) under Project Detail Page:

```
- If the connected wallet is the project recipient and the cycle is still active, show an "Edit Project" button that opens the same form pre-filled with current values and calls updateProject.
```

carletex:
> Should we add to the prompt the ability to edit your own projects? 

This is fair! but I don't think it adds (conceptually) anything, and it's an implementation detail. And maybe it's not desirable (or only for some fields, or only when the project hasn't received any funding, etc)

The way I see it, these prompts are meant to be starting points.


Merging!

PR #383, opened 2026-04-17
Improve token vesting prompt

damianmarti:
Tested with Sonnet 4.6, Composer 2, and GPT 5.4.

The Solidity code works well and looks good, but the frontend depends on the model used. Added some instructions so the frontend features do not depend on the model.

Added shorter vesting period on the frontend for better manual testing, too.


rin-st (approved):
Tested this just with Opus and it worked well. Merging, thank you!

PR #384, opened 2026-04-20
fix impermanent loss price ratio derivation in guide

Pabl0cks:
## Summary
- correct the price ratio definition from `x/y` to `y/x` in the impermanent loss derivation
- align the initial and post-price-change equations with the later reserve and valuation formulas

## Why
The guide later uses `x₁ = √(k/p₁)`, `y₁ = √(k × p₁)`, and `Value_LP = y₁ + (x₁ × p₁)`, which all imply that `p` is the price of token `x` in units of token `y` (`p = y/x`).


rin-st (approved):
Lgtm!

PR #389, opened 2026-06-08
Add token-abuse protection to the AI assistant (open it to registered builders)

technophile-04:
Adds token-abuse protection to the per-challenge AI assistant and opens it up from admin-only to all registered builders.

### What it does

- Per-wallet **daily token budget** (default 1M tokens/day), tracked in Postgres — checked before each request, and the real usage is charged after the stream finishes in `onFinish`.
- Per-response **output cap** (default 5k tokens) so a single answer can't blow up the cost.
- Both tunable via env (`CHAT_DAILY_TOKEN_BUDGET`, `CHAT_MAX_OUTPUT_TOKENS`) — no redeploy to retune.
- The assistant now shows to any **connected + registered** builder (was admin-only); the API only verifies the wallet is registered.
- When someone hits the limit, the chat shows a friendly card that nudges them to set the challenge up locally (Checkpoint 0 → `/start`).

### Why these choices

- **Tokens, not message count** — tokens are the actual unit of cost. A one-word "thanks" and a long debugging answer shouldn't count the same.
- **Postgres, not Redis/Upstash** — we already run Postgres + Drizzle, so no new service or credential. The counter is an atomic upsert (`ON CONFLICT DO UPDATE ... + n`) so concurrent requests can't race past the budget.
- **Identity is the connected wallet from the request body.** SRE only establishes a SIWE session on `/admin`, so normal builders have no server-verified identity. We gate on the address being a registered builder and keep it zero-friction (no sign-in step). The trade-off is below.
- **Fixed UTC-day window** — one atomic counter per (wallet, day). A rolling 24h window would need timestamped usage for marginal benefit.

### Known limitation (accepted for now)

> Identity is the wallet address the client sends — not proof of ownership. A caller could send any registered address (their own, someone else's to burn that wallet's budget, or rotate throwaway registrations). The model is cheap and the output cap bounds per-request spend, so this still stops casual abuse. Hardening is a TODO below.

Talked with carlos regarding this in today's meet and we decided for now lets ship it with impersonation, most of the people won't take go the line to exploit it too much 

### How to test

```bash
# packages/nextjs, with a local DB + an AI provider key set
CHAT_DAILY_TOKEN_BUDGET=1 yarn start
# connect a registered wallet, open an AI-ready challenge, send two messages:
#   msg 1 → streams a reply (POST /api/chat 200)
#   msg 2 → 429, and the chat shows the "take it local" limit card
```

### Deferred

- [x] Conversation logging + observability (companion PR)
- [ ] EIP-712 "unlock" signature so the server can verify wallet ownership without a full SIWE sign-in
- [ ] Model upgrade `qwen3.5-27b` → `qwen3.6-27b` (near-identical price)


packages/nextjs/utils/ai/rate-limit.ts
```diff
@@ -0,0 +1,23 @@
+import { addDailyTokenUsage, getDailyTokenTotal } from "~~/services/database/repositories/chatTokenUsage";
+
+// Token-abuse limits for the AI chat assistant
+export const DAILY_TOKEN_BUDGET = Number(process.env.CHAT_DAILY_TOKEN_BUDGET ?? 1_000_000);
+export const MAX_OUTPUT_TOKENS = Number(process.env.CHAT_MAX_OUTPUT_TOKENS ?? 5_000);
```
rin-st: **Brittle env parsing — a misconfig silently breaks the rate-limit gate** — `utils/ai/rate-limit.ts:4-5`

  `Number(process.env.X ?? default)` — `??` only guards null/undefined, not `""`. So:

  - `CHAT_DAILY_TOKEN_BUDGET=""` (empty value set in Vercel) → `Number("")` = `0` → `used >= 0` is always true → **chat disabled for everyone (429)**.
  - `CHAT_DAILY_TOKEN_BUDGET=abc` (typo) → `NaN` → `used >= NaN` is always false → **limit silently off, unlimited spend**.

  Same applies to `CHAT_MAX_OUTPUT_TOKENS` (non-numeric/`0` → the SDK throws `InvalidArgumentError` per request → 500s).

  Parse + validate once at module load, fall back to the default on invalid input:

  ```ts
  function envInt(name: string, fallback: number): number {
    const n = Number(process.env[name]);
    return Number.isFinite(n) && n > 0 ? n : fallback;
  }
  export const DAILY_TOKEN_BUDGET = envInt("CHAT_DAILY_TOKEN_BUDGET", 1_000_000);
  export const MAX_OUTPUT_TOKENS = envInt("CHAT_MAX_OUTPUT_TOKENS", 5_000);
  ```

  Side note: the `.env.example` "tune in Vercel without a redeploy" comment is optimistic — these are module-level consts read at cold start, so a warm instance keeps the old value until it recycles.

rin-st (approved):
Added one nitpick from AI review, since the typo in env var could lead to unlimited spend. 

Other than that looking and working great!

packages/nextjs/utils/ai/rate-limit.ts
technophile-04: good catch on the `??` gap! fixed at pushed at small commit https://github.com/BuidlGuidl/SpeedRunEthereum-v2/commit/504fbaca1f83102d00e5d31e26d06db9b65c819c , went with `Number(x) || fallback` instead of the validating helper though. these vars are ours to set, so went with keeping the guard to the one case that actually bites than defend every shape a bad value could take (empty / NaN / negative / …). I remember this [Thorsten Ball essay](https://registerspill.thorstenball.com/p/joy-and-curiosity-81) where he mentioned, agents tend to write defensive "scared mouse" code, a what-if on every line, and it compounds: the next reader (human or agent) trusts the codebase a bit less and piles on more. trying to resist that and keep it confident

technophile-04:
Thanks Rinat! Will merge this when I get the user message logs pr ready...so that on the day of launch to normal users, we have the logs setup

packages/nextjs/utils/ai/rate-limit.ts
rin-st: > [Thorsten Ball essay](https://registerspill.thorstenball.com/p/joy-and-curiosity-81) where he mentioned, agents tend to write defensive "scared mouse" code, a what-if on every line, and it compounds

Good point, thanks for the essay!

technophile-04:
Merged the [session chat logging PR](https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/391). Configured the environment on for easy testing and also to test if the migration works. 

Test at https://speedrunethereum-v2-git-ai-chat-logging-ra-07f511-buidlguidldao.vercel.app/ 

For the token-abuse, we have capped 1M token (input + output)/ day. The daily usage for all users resets at UTC time  `00:00:00 UTC` for simplicity

### Environment config: 

#### Neon: 

> Before merging this branch we need to run `yarn drizzle-kit migrate` on main branch

Configured the `ai-chat-public` branch (branched out from today's main) and then tried running the `yarn drizzle-kit migrate` on that branch and everthing works so green flag for running it on `main` as well. 

#### Vercel: 

- Openrouter Key: I've updated all the environments (preview + prod) to use the BG openrouter API key. The model we are using is `qwen/qwen3.5-27b`   

- For this PR branch we are using the `ai-chat-public` neon database

- Configured `CHAT_DAILY_TOKEN_BUDGET=1M` (for prod) and for this branch kept it at 20K so that we can teset and `CHAT_MAX_OUTPUT_TOKENS=500`


Created https://sre-logging-flow.vercel.app/ for the overview of the review. 

--- 

### Some probable UI tweaks before launching: 

- We need to change the Icon? Currently, we have sparkle

rin-st:
Thank you for details! Working great!

nit, not related to this PR. Is it still named simple-nft in db?
Screenshot 2026-06-11 at 14 55 59

rin-st (approved):

technophile-04:
> nit, not related to this PR. Is it still named simple-nft in db? 

ohh no, its just my claude probably picked it from memory for coming with example. I just re-confirmed by checking the DB as well as concepts.yaml if we leak that ref and there was no mention of it. Asked the LLM and it also mention it just fabricated as an example

rin-st:
> ohh no, its just my claude probably picked it from memory for coming with example. I just re-confirmed by checking the DB as well as concepts.yaml if we leak that ref and there was no mention of it. Asked the LLM and it also mention it just fabricated as an example

Got it, thanks!

carletex:
This is looking pretty good!

q: How hard would it be to keep the conversation/session when the user refreshes or goes back for a given challenge? only one at a time (per challenge), so you can either continue a convo or start over (with the button already there)

carletex:
Pushed a lil tweak that show this tooltip for a few seconds:
image

There is a lot of things going on (Submit challenge button, dark mode, AI teacher), so I think it's good to show it.

> q: How hard would it be to keep the conversation/session when the user refreshes or goes back for a given challenge? only one at a time (per challenge), so you can either continue a convo or start over (with the button already there)

We are already storing in database, but it'd be tricky because of impersonation. We could fix this by:
- Use localstorage: might be enough
- make people sign (as admins). I think Shiv already mentioned this somewhere.

In any case, we can leave it for another PR. Let's discuss this week.

Merging! (and running the migrations)

PR #391, opened 2026-06-10
AI chat: conversation logging (session-per-row jsonb)

technophile-04:
## Summary

Conversation logging for the per-challenge AI chat assistant. One chat session = one row instead of per-message as row (thanks Carlos for the idea), the whole conversation stored as a single `jsonb` document.  

## Concretely

- New `chat_conversations` table (migration `0016`): one row per session, `messages jsonb` holds the assembled `UIMessage[]`, plus `messageCount`, `createdAt`, `updatedAt`. `id` is the client-minted `conversationId` (PK), not server-defaulted.
- `upsertChatConversation` repo: `INSERT ... ON CONFLICT (id) DO UPDATE` — the first turn inserts, every later turn rewrites the same row with the grown transcript (no per-message diffing).
- `/api/chat` persists the transcript in `toUIMessageStreamResponse`'s `onFinish` (the *full* `UIMessage[]`), separate from the token-charging `onFinish` on `streamText`. `result.consumeStream()` (no `await`) so the row still saves if the builder closes the tab mid-answer.
- `ChatWidget` mints a `conversationId` (`crypto.randomUUID()`) at mount and on reset, and sends it in the request body via a ref (so the memoized transport reads the current id). A new id = a new session row.

Also just created a companion html doc to easily digest the things in code: https://sre-logging-flow.vercel.app

## Why session-per-row (not message-per-row)

On ai-sdk v6 a `UIMessage` is a structured object (`parts[]` + `metadata`), and `toUIMessageStreamResponse` hands back the whole assembled array — so storing the array whole is the SDK's own persistence shape. Per-turn detail lives in Opik when it's enabled, so a per-message table would duplicate it. No digest/insight columns: aggregation reads these blobs later into its own store keyed by `conversationId`, keeping this table a pure append-only log. `chat_token_usage` (the gate) stays its own table — different grain `(userAddress, day)`.

## How to test

```bash
git checkout ai-chat-conversation-logging
cd packages/nextjs
docker compose up -d
yarn drizzle-kit push   # local migrations journal is behind from prior `push` syncs; `push` adds the table. `migrate` runs clean on prod's intact journal.
yarn start
```

Open a challenge, connect a registered wallet, open the assistant, send a couple of messages. A `chat_conversations` row appears keyed by the client `conversationId`; `messageCount` grows each turn (same row). Hit reset, send again — it lands in a new row.

Write path also verified directly against local postgres: first upsert inserts, second upsert with the same id rewrites in place (`createdAt` unchanged, `messageCount` and `updatedAt` advance, exactly one row).

## Deferred

- Opik wiring — zero-code OpenRouter toggle, left off until there's a specific question to watch.
- Digest / insight aggregation into a separate store keyed by `conversationId`.

## Why store the whole message

We log the full `UIMessage[]` per session rather than a slimmed role+text projection. The assistant has no tool calls, so the parts are just text + reasoning, and the reasoning trace is the most useful signal for prompt tuning (catching where the tutor leaks answers). Storage is negligible — a few KB of TOAST-compressed jsonb per session, and the one redundancy (reasoning duplicated in `providerMetadata`) compresses away. Any slim view is a read-time projection in the later digest step, not a lossy write-time transform.

### DIFF
--- a/packages/nextjs/app/api/chat/route.ts
+++ b/packages/nextjs/app/api/chat/route.ts
@@ -28,7 +28,7 @@ export async function POST(req: NextRequest) {
     return NextResponse.json({ error: "Register as a builder to use the assistant." }, { status: 403 });
   }
 
-  // conversationId is the client-minted session id; it keys the persisted transcript row (ADR 0004).
+  // conversationId keys the persisted transcript row.
   if (!messages || !challengeId || typeof conversationId !== "string") {
     return NextResponse.json({ error: "Missing required fields" }, { status: 400 });
   }
--- a/packages/nextjs/app/api/chat/route.ts
+++ b/packages/nextjs/app/api/chat/route.ts
@@ -9,9 +9,7 @@ import { buildChatSystemPrompt } from "~~/utils/ai/system-prompt";
 
 export const maxDuration = 60;
 
-// Server-side id for each new assistant message. The SDK's generator (prefixed, collision-resistant)
-// is the canonical persistence choice; the id is streamed to the client so its copy and the stored
-// copy match. Distinct from conversationId (the row PK / opik session id), which stays a uuid.
+// Stable ids for persisted assistant messages (streamed to the client so both copies match).
 const newMessageId = createIdGenerator({ prefix: "msg", size: 16 });
 
 export async function POST(req: NextRequest) {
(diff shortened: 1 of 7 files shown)


rin-st (approved):
Working great!

technophile-04:
Thanks Rinat! Merging this 🙌

PR #390, opened 2026-06-08
Add AI dApp workflow guide with on-page navigation

Pabl0cks:
The idea of this article is to cover the entire dApp creation workflow, with a short introduction to how AI helps in each phase. It will be complemented by deep dives covering 1–3 steps from the workflow at a time, for a total of 4–5 articles.

This first article will serve as a high-level overview and as an index once the rest of the articles are published.

This PR is bigger than usual because I’ve not only added the article, but also included a couple of features to make it nicer to consume:

-  A navigation side menu, similar to the challenges approach, but with a nicer UI (I hope!)
-  The ability for users to open article images at full size.

## Summary

Adds a new long-form guide and the on-page navigation needed to read it comfortably:

- **New guide:** "How to build a dApp on Ethereum with an AI workflow" (markdown + thumbnail, inline diagrams, a build-prompts clip, and the AI Tutor demo video embed).
- **TOC sidebar for guides:** opt-in via a `showNavigation` frontmatter flag (default off). When enabled, the guide hero and article share a two-column grid with a sticky, vertical-rule table of contents that scroll-spies the active section. Desktop only. h2 headings are extracted server-side and matching ids are stamped on the rendered headings so anchors resolve; fenced code blocks are skipped so example snippets aren't indexed as sections.
- **Anchor-id fix:** heading ids were built from two different inputs (the rendered `<h2>` hashed `String(children)`, turning inline markup into `"[object Object]"`, while the TOC hashed the raw markdown), so anchors and scroll-spy broke for headings containing code, bold, or links. Ids are now generated from flattened plain text on both sides.
- **Click-to-zoom lightbox for guide images:** markdown images in any guide now render as clickable thumbnails that open a full-screen overlay showing the image at its original size (click the overlay, the ✕ button, or press Escape to close; body scroll is locked while open). Theme-aware styling for light/dark mode. Adds `@types/react-dom` (devDependency) for the `createPortal` import.
- **Raw HTML + hydration fix:** guide markdown is now processed with `rehype-raw` so raw HTML (tables, video embeds via a responsive `iframe` mapping) renders properly. A small rehype plugin strips whitespace-only text nodes from table structure elements (`table`/`thead`/`tbody`/`tr`/...) — `rehype-raw` keeps the indentation text nodes from the markdown source, and React 19 rejects them inside tables, reporting 4 hydration errors on every page load.

## How to test

1. `cd packages/nextjs && yarn dev`
[...]

PR #392, opened 2026-06-12
Redesign challenge page layout

Pabl0cks:
Tried to redesign the challenge layout to be more inline with the new version of the guides layout from #390 

- In the hero i tried to have clean layout with title + skills and challenge compact image in the right, used white background to have nice contrast with the greens from the challenge images. 
Having the skills at the top gives us a nice summary of what's the challenge about, and it's cleaner to add the image to the right.
- Deleted the AI assisted / guided paragraph, and added a sentence in the AI Teacher chat on its empty state, with a link to the AI-Guided section of the readme (if exists). 
   - Maybe we should add some kind of text to the AI Teacher button to make it more visible so whoever wants assistance, finds it easily.
- Increased the size of the font of the empty state chat because it felt a bit too small
- Pushed the content to the left if the user opens the AI Teacher panel

## Summary

Reworks the per-challenge page into a more structured, scannable layout:

- **Hero** with a clean "Challenge: {name}" title (no emojis), the "Skills you'll gain" list in a single column, and the challenge preview image full-size beside them. White background in light mode so the pixel-art images read as designed; dark teal in dark mode.
- **Facts bar** — skill level, time to complete, completed-by count, and helpful links pulled into a single card that overlaps the hero.
- **Content card** — the full readme body (intro paragraph included), "View on GitHub", related guides, and submit/register actions flow inside one card.
- **Sidebar** ("On this page") restyled with a continuous left rule and sticky positioning; it reserves its column on large screens even when empty so the content doesn't jump, and long TOCs get a thin hover-only scrollbar instead of the OS default.
- **AI chat widget**: below 2xl the open panel overlays the content (like mobile); at 2xl+ the page reserves the panel's width (`data-challenge-chat-open` + `.challenge-page-shell` in `globals.css`). Copy/sizing polish in the header and empty state; the empty state's "AI-guided section" link is derived from the readme's extracted headings (same ids the sidebar uses) and hidden when the section is absent.

**Chat gating is intentionally unchanged (admin-only).** Opening the assistant to users + rate limiting is handled separately in #389, so this PR deliberately leaves `api/chat/route.ts` and the widget's auth gate as they are on `main` to avoid overlapping with that work.

## How to test

1. `cd packages/nextjs && yarn start`
2. Open any challenge page (e.g. an AI-ready one) and check:
   - Hero shows the "Challenge: {name}" title, skills list, and preview image (light + dark mode).
   - Facts bar renders and wraps correctly on mobile and desktop.
   - "On this page" sidebar scroll-spy still works; on a long TOC (e.g. stablecoins) the scrollbar only appears on hover.
3. As an admin, open the AI chat widget: at ≥1536px wide the content shifts left to make room; below that the panel overlays the page. No horizontal scrollbar at any width.

## Notes for reviewers

- Touches the same two chat files as #389 (`api/chat/route.ts`, `ChatWidget/index.tsx`) but only the widget's visual bits — the auth gate and route are identical to `main`. Whichever of the two PRs merges second, the chat-related overlap should be trivial.
- `ChallengeHeader` now takes `children` (the challenge body + actions) instead of rendering the AI-ready blurb; the `isAiReady` prop was removed from the header. Known follow-up: non-admin users currently see no "AI-ready / run `/start`" messaging on AI-ready challenges (the old info card was removed and the chat widget is admin-gated) — to be addressed here or after #389.
- `splitChallengeReadme`'s `headerImageMdx` is no longer rendered (the hero renders the title directly); the readme body is rendered in full, with the readme H1 stripped only when followed by the header image, per current readme convention.

### DIFF
--- a/packages/nextjs/app/challenge/[challengeId]/_components/ChallengeHeader.tsx
+++ b/packages/nextjs/app/challenge/[challengeId]/_components/ChallengeHeader.tsx
@@ -1,14 +1,10 @@
 import type { ReactNode } from "react";
-import { MDXRemote } from "next-mdx-remote/rsc";
-import rehypeRaw from "rehype-raw";
-import remarkGfm from "remark-gfm";
-import { CheckCircleIcon, ClockIcon, LinkIcon, UserGroupIcon } from "@heroicons/react/24/outline";
+import { ClockIcon, LinkIcon, UserGroupIcon } from "@heroicons/react/24/outline";
 import SkillLevelIcon from "~~/app/_components/SkillLevelIcon";
 import { SkillLevel } from "~~/utils/challenges";
 
 type Props = {
   title?: string;
-  skills?: string[];
   skillLevel?: SkillLevel;
   timeToComplete?: string;
   helpfulLinks?: { text: string; url?: string }[];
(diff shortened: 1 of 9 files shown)


carletex:
Thanks Pablo, looking great! Made a few tweaks:

- Fixed the " Ready to submit your challenge?" banner z-index. It was going under the "skills" one (noticeable on mobile). Also made the banner a bit smaller on mobile
- Added back the AI banner (I think we'll want to remove the inlined texxt from every challenge README... and add it into that top one)
- Reverted back the old width
- "Pushed the content to the left if the user opens the AI Teacher panel" this looked a bit weird. Removed it. We can tweak the UX if needed.

MErging!

PR #394, opened 2026-06-25
Update bonding curve prompt

rin-st:
Prompt was already good, I just tweaked ui theme, added grumpy reviewer and asked to use separate react components for every feature

Tested with Opus. UI was great, and everything worked


damianmarti (approved):
Tested without and with the new changes. 

The new code is better organized and forced the review with grumpy-carlos.

Thanks!

rin-st:
Thanks for the review!

PR #395, opened 2026-06-25
Update streaming payments prompt

rin-st:
- ui theme tweak
- time-based blocks (1 per second)
- more details in frontend spec
- grumpy reviewer

Tested with Opus. UI was good and everything worked


damianmarti (approved):
Tested with Sonnet. Working great!

I think we should add "Implement a React component for each feature." to **Required Views** to avoid having a lot of code into the streaming page (like you added to bonding curve).

Thanks!

rin-st:
> I think we should add "Implement a React component for each feature." to Required Views to avoid having a lot of code into the streaming page (like you added to bonding curve).

Updated, thanks! Merging this

PR #399, opened 2026-07-07
Add homepage H1 and self-referencing canonical URLs

denalimarsh:
## Summary

Two baseline crawl/SEO gaps from the audit:

- **Homepage had no `<h1>`.** `Hero.tsx` rendered the tagline as a `<p>`. Promoted it to a semantic `<h1>`.
- **No self-referencing canonicals.** `getMetadata` emitted no canonical (only `/learn-solidity` set one manually). Added an optional `path` param to `getMetadata` that sets `alternates: { canonical: path }`, resolved against the existing `metadataBase`, and wired it through the content pages.

`/learn-solidity` keeps its existing canonical. Dev-only pages (`/debug`, `/blockexplorer`) are excluded.

## How to test

2. View source on `/`: exactly one `<h1>`, and `<link rel="canonical" href=".../">`.
3. Open a challenge and a guide: each has a `<link rel="canonical">` pointing at its own URL.

### DIFF
--- a/packages/nextjs/app/_components/Hero.tsx
+++ b/packages/nextjs/app/_components/Hero.tsx
@@ -12,7 +12,7 @@ export const Hero = () => {
           <HeroDiamond className="h-12 w-12" />
         </div>
 
-        <h1 className="text-center mb-5  dark:text-gray-200">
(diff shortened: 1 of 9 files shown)


carletex:
Looking good. Thanks!

PR #405, opened 2026-08-04
Add video guide: how a full-stack Ethereum app is structured (Devcon SEA talk)

denalimarsh:
## What this adds

Second guide in the video series started in #404, from Austin's Devcon SEA talk: video embedded, timestamped outline, edited transcript.

`/guides/how-a-full-stack-ethereum-app-is-structured`

## Note on the source

This video is on the Ethereum Foundation channel rather than Austin's, so it's worth a nod from EF before merge. Happy to adjust the credit line or framing.

## Why

The most complete architecture walkthrough on video: the two packages, how artifacts reach the frontend, burner wallets, Debug Contracts, then the curriculum with a staking contract written live.

Same gap as #404. Across 50 developer questions, YouTube was cited 0 times out of 546 sources; speedrunethereum.com was cited 12. Two of those questions ask how to structure a full Ethereum app and what the cleanest layout for contracts, frontend and tests is. Neither surfaces Scaffold-ETH 2 today.

## About the transcript

Edited down rather than verbatim, unlike #404. That video was 8 minutes of tight demo. This is 100 minutes and ~15,000 caption words, including a long stretch debugging a testnet RPC before moving to Optimism.

Every section is still covered and anchored in order. What's cut is dead air, repeated setup and debugging detours. Mangled terms were checked against docs.scaffoldeth.io. Wording and asides left as spoken.

Say the word if you'd rather have it verbatim for consistency, since this sets the pattern for the remaining videos.

## Notes for review

- 17 `h2` sections, 17 `h3` sub-points. `h3` only gets an anchor id and a sidebar entry once #409 lands, which is why this depends on it. Flattening everything to `h2`, as #404 does, works at 8 minutes but gives a 34 item table of contents at 100.
- Headings lead with words, time range at the end.
- Staking challenge linked as **Crowdfunding**, its current name, noting the talk calls it Decentralized Staking.
- Prose uses "Speedrun Ethereum"; camelCase only inside the video title.
- Includes `llms.txt` and `sitemap.xml` entries.
- Thumbnail is the video's own.

## Files

- `packages/nextjs/guides/how-a-full-stack-ethereum-app-is-structured.md`
- `packages/nextjs/public/assets/guides/how-a-full-stack-ethereum-app-is-structured.jpg`
- `packages/nextjs/public/llms.txt`
- `packages/nextjs/public/sitemap.xml`

No code changes.

PR #404, opened 2026-08-03
Add video guide: build an Ethereum app in 8 minutes, with timestamped transcript

denalimarsh:
## What this adds

A guide built around one of Austin's videos: the video embedded, a timestamped outline under it, and the full transcript with matching section headers, so every part of the video is readable and linkable as text.

`/guides/build-an-ethereum-app-in-8-minutes`

## First of a set

This is the first of Austin's videos to get this treatment, and it is meant as the template rather than a one-off. Several more videos are worth activating the same way, so the outline structure, the section granularity and the level of transcript cleanup are all open for discussion now: whatever we agree on here, the rest will follow. Much easier to change the model on one page than on six.

## Why

The video covers the whole loop in eight minutes: `create-eth`, the contract, Debug Contracts, hooks and components, deploy to Base, ship the frontend. It is one of the clearest introductions to the current stack we have, and none of it is reachable by search engines or AI assistants, because none of them watch video.

In a measurement run across 50 developer questions, YouTube was not cited once across 546 sources, while speedrunethereum.com was cited 12 times. Putting the content on a page here makes it citable.

## About the transcript

Cleaned from YouTube's auto-captions, which are rough. Edits are limited to punctuation, capitalisation, and terms the captions mangled:

- "MPX create E latest" → `npx create-eth@latest`
- "CDN to" → `cd` into
- "Versell YOLO prod" → `yarn vercel:yolo --prod`
- "use scaffold read contract" → `useScaffoldReadContract`
- "Vitalic" → Vitalik

Every command was checked against docs.scaffoldeth.io rather than trusted from the captions. Wording, asides and jokes are left as spoken.

## Notes for review

- First guide to embed a video. No new component needed: the `iframe` mapping in `services/guides.ts` already wraps embeds responsively.
- Transcript sections are `h2`. At 8 minutes the outline is short enough that a flat table of contents reads well as a scrubber.
- Thumbnail is the video's own YouTube thumbnail.
- @austintgriffith: this republishes your talk, so it needs your OK on the framing and the byline before merge.

## Files

- `packages/nextjs/guides/build-an-ethereum-app-in-8-minutes.md`
- `packages/nextjs/public/assets/guides/build-an-ethereum-app-in-8-minutes.jpg`

No code changes.

Unrelated, spotted while working: an existing guide links to `/challenge/svg-nft`, which is not in the `ChallengeId` enum. Happy to fix separately.

PR #406, opened 2026-08-04
Add video guide: deploy your first smart contract to an L2 (Arbitrum)

denalimarsh:
## What this adds

Third guide in the video series (#404, #405), from Austin and Elliot's Arbitrum walkthrough: video embedded, timestamped outline, edited transcript.

`/guides/deploy-your-first-contract-to-an-l2`

## Why

Targets "how do I deploy my first contract that I can interact with it", a Building prompt where Scaffold-ETH 2 doesn't currently surface. The first half is a clean from-zero path: scaffold, tinker locally, fund a deployer, deploy for 25 cents. The extensions pitch is the part I'd most want indexed, since "your starter kit stays state of the art" answers the objection that sinks most forked templates.

## Notes for review

- This one is a dialogue. Elliot built the Arbitrum extension and explains the address table precompile and forced transactions, so speaker labels run throughout. **Please check I have his name right**, I inferred the surname rather than confirming it.
- Roughly half the runtime is Arbitrum-specific rather than Scaffold-ETH. Kept because it's genuinely good, and titled "to an L2" so the generalizable claim leads.
- Two internal links point at the guides in #404 and #405, so they 404 until those merge.
- Transcript edited for length, same as #405. 2,920 words from ~4,700 caption words.
- Includes `llms.txt` and `sitemap.xml` entries. Thumbnail is the video's own.

## Files

Four, same shape as #405. No code changes.


denalimarsh:
@Pabl0cks @carletex @austintgriffith review ping. Content only, no app code. @austintgriffith I wrote the extension kit steps against the current flow, not the one in the video, so worth a check.

PR #408, opened 2026-08-04
Add video guide: getting front-run on purpose, and what the mempool exposes

denalimarsh:
## What this adds

Fifth guide in the video series (#404, #405, #406, and the Columbia one alongside this), from the Bow Tie Friday stream where Austin ships a deliberately front-runnable contract to mainnet: video embedded, timestamped outline, edited transcript.

`/guides/getting-front-run-on-purpose`

## Why

Testing and simulation is the weakest category we track, and this video has something no explainer can offer: a result. The same contract with the same bait is ignored on Optimism and front-run almost immediately on Ethereum mainnet, for about a hundred dollars, and the block explorer shows the attacker's transaction landing twelve seconds ahead with the value splitting between the attacker and the builder.

The sequencer contrast is the part I would most want indexed. Nothing in the code changed between the two networks, so it isolates what a public mempool actually exposes.

## Notes for review

- **Positioned as the empirical companion to two guides we already have**, not a third explainer. `/guides/commit-reveal-scheme` and `/guides/front-running-mev-mitigation` cover the defensive implementations, and this page links to both rather than repeating them. Worth a check that the framing reads right to you.
- The contract is deliberately vulnerable and Austin says so on camera ("all sorts of holes in this"). The intro, an FAQ, and the summary all say plainly that it is not a pattern to copy. That felt necessary given it will be read out of context.
- Livestream, so a real amount of runtime is deployment waiting and getting the stream link out. Compressed to one bracketed line. 1,827 words from ~6,100 caption words, the heaviest edit in the series so far.
- Includes `llms.txt` and `sitemap.xml` entries. Thumbnail is the video's own.

## Files

Four, same shape as #406. No code changes.


denalimarsh:
@Pabl0cks @carletex @austintgriffith review ping. Content only, no app code. @austintgriffith kept the framing descriptive rather than how-to, shout if you would pitch it differently.

PR #409, opened 2026-08-04
feat: give h3 subheadings anchors and TOC entries in guides

denalimarsh:
## What

`h3` headings in guides are now indexable: they get an anchor id and appear in the "On this page" sidebar, indented under their `h2`.

Previously only `h2` did, so any `###` was unlinkable and invisible in the nav. On long guides that meant either losing the subsections or promoting them all to `h2` and ending up with a flat 30-plus item sidebar.

## How

- `extractHeadings` takes a `maxLevel`, defaulting to 2. Guides pass 3.
- `h3` gets an id in `mdxComponents`, same as `h2`.
- The sidebar indents level-3 entries.

Challenge pages use the default and are unchanged. Existing anchors are unchanged.

## How to test

`yarn dev`, open a guide with `###` headings: indented sidebar entries, scroll-spy works, clicking updates the hash. Challenge pages look the same as before.


denalimarsh:
Worth noting the scope, since the description undersells it: this is not just for new long guides. 20 of the 24 guides already published contain `###` subheadings, 231 in total, and none of them currently get an anchor id or a sidebar entry. This makes all of those deep-linkable and adds them to the table of contents without touching any guide content.

denalimarsh:
@Pabl0cks @carletex ping. Only code change in the batch, +19/-6 across three files. Extends anchors and TOC entries from h2 to h3, existing h2 anchors unchanged.

PR #401, opened 2026-07-21
Self-identify guides with breadcrumb and TechArticle structured data

denalimarsh:
## Summary

Follow-up to #398, extending the same structured data approach to the guides.

- **Guides do not name Speedrun Ethereum anywhere in the page content.** The header logo is client-rendered, so crawlers that skip JavaScript never see the brand on guide pages. Added a small "Speedrun Ethereum › Guides" breadcrumb above the guide h1. Both segments are working links (home and the guides index).
- **Guides emit no page-level structured data**, only FAQ schema when the frontmatter has `faqs`. Added `getGuideStructuredData()` in `utils/structuredData.ts`: a `TechArticle` node (headline, description, image, datePublished, `mainEntityOfPage`, publisher BuidlGuidl, and `isPartOf` pointing at the shared `#course` entity from #398) plus a `BreadcrumbList` matching the visible breadcrumb. Merged with the existing FAQ schema into a single JSON-LD script.

Canonicals were already handled in #399, so no changes there. No layout changes beyond the one breadcrumb line.

## How to test

1. `cd packages/nextjs && yarn dev`
2. Open any guide, e.g. `/guides/how-to-create-erc20`: a "Speedrun Ethereum › Guides" breadcrumb renders above the title and both links resolve.
3. View source: one `ld+json` script containing `TechArticle` and `BreadcrumbList` (plus `FAQPage` on `/guides/how-to-build-dapp-ethereum-ai-workflow`, the one guide with `faqs`).
4. `yarn check-types` passes.

PR #398, opened 2026-07-07
Add Course and ItemList structured data to homepage and challenges

denalimarsh:
## Summary

Only guide pages emit JSON-LD today (`FAQPage`, in `app/guides/[slug]/page.tsx`). The homepage and challenge pages, which are the learning product itself, emit none, so parsers can't recognize "Speedrun Ethereum" as a structured course with an ordered curriculum.

This adds `schema.org` structured data via `<script type="application/ld+json">`, reusing the existing guides pattern:

- **Homepage:** a `Course` node (provider: BuidlGuidl) plus an `ItemList` of the challenges in curriculum order, built from the live `challenges` data and sorted by `sortOrder` (disabled challenges excluded).
- **Challenge pages:** a per-challenge `Course` node marked `isPartOf` the overall Speedrun Ethereum course, using the same title/description as the page metadata.
- New helper `utils/structuredData.ts` keeps the markup out of the page components.
- The existing guide FAQ schema is unchanged.

For example, `/challenge/dex` returns the following structured data:
```
{
  "@context": "https://schema.org",
  "@type": "Course",
  "name": "Build a DEX",
  "description": "Step-by-step tutorial to build your own DEX with liquidity pools, token swapping, and automated market making in Solidity.",
  "url": "https://speedrunethereum.com/challenge/dex",
  "provider": {
    "@type": "Organization",
    "name": "BuidlGuidl",
    "url": "https://buidlguidl.com"
  },
  "isPartOf": {
    "@type": "Course",
    "name": "Speedrun Ethereum",
    "url": "https://speedrunethereum.com"
  }
}
```

## How to test

1. `cd packages/nextjs && yarn dev`
2. View source on `/` and confirm two `application/ld+json` blocks: a `Course` and an `ItemList` whose items match the real challenge order.
3. Open any `/challenge/<id>` and confirm a `Course` block with `isPartOf` pointing back to the curriculum.
4. Paste either into the [schema.org validator](https://validator.schema.org/) and confirm it parses with no errors.

PR #411, opened 2026-08-05
Add structured data to /learn-solidity, /start and /build-prompts

denalimarsh:
## Summary

Follow-up to #398 and #401, extending the same entity graph to the three top-level pages that still emit nothing.

- **Three pages emit no structured data at all.** The homepage has `Course` + `ItemList` (#398), guides have `TechArticle` + `BreadcrumbList` (#401), and challenge pages have a `Course` node, but `/learn-solidity`, `/start` and `/build-prompts` emit no `ld+json` whatsoever. `/learn-solidity` is the strongest `Course` candidate on the site and was missed in #398.
- **Added `getTopLevelPageStructuredData()`** in `utils/structuredData.ts`, and wired it through the three pages:

| Page | Emits |
| --- | --- |
| `/learn-solidity` | `WebPage` + `BreadcrumbList` + `FAQPage` |
| `/start` | `WebPage` + `BreadcrumbList` |
| `/build-prompts` | `CollectionPage` + `BreadcrumbList` |

**On not minting a second `Course`.** The obvious move on `/learn-solidity` is to emit its own `Course` node, but that would give the site two competing course entities and leave the page weaker than it is now. Instead each page is a `WebPage` (or `CollectionPage`, reusing the `getGuidesIndexStructuredData` shape) whose `isPartOf` and `about` both point at the existing `#course` entity from #398. The site keeps exactly one course node.

The `FAQPage` on `/learn-solidity` describes the five answers already rendered on the page. No copy changes here.

Each page's title and description are hoisted into constants shared by the metadata and the structured data, so the two cannot drift. `/learn-solidity` also moves from its hand-written `alternates.canonical` to the `path` argument added in #399, which produces the same canonical URL.

## How to test

2. View source on `/learn-solidity`: one `ld+json` script containing `WebPage`, `BreadcrumbList` and `FAQPage`, with the five existing FAQ answers.
3. Same on `/start` (`WebPage` + `BreadcrumbList`) and `/build-prompts` (`CollectionPage` + `BreadcrumbList`).
4. Confirm `"@id": "https://speedrunethereum.com/#course"` appears as an `isPartOf`/`about` reference on all three, and that `"@type": "Course"` with that `@id` is still only declared on the homepage.
5. Canonicals are unchanged: `/learn-solidity` still emits `<link rel="canonical" href=".../learn-solidity">`.
6. `yarn check-types` passes.


denalimarsh:

PR #400, opened 2026-07-07
Self-identify Speedrun Ethereum in challenge, guide, and landing copy

denalimarsh:
Names Speedrun Ethereum as the source across metadata and existing on-page copy, so AI assistants that already surface these pages credit the brand.

## Why
In AI-visibility testing, LLMs (Claude) already cite Speedrun Ethereum pages when answering "how to learn / how to build X on Ethereum" questions - but often name other resources as the answer, because the pages don't clearly identify Speedrun Ethereum as the source. This adds the brand where it's currently missing, with no layout or design changes.

## What changed
- **Challenge meta descriptions** (`utils/challenges.ts`): each now names Speedrun Ethereum and leads with the build + "free / hands-on" framing (head metadata only; no visible page change).
- **Guide closing CTAs** (14 guides): the existing "Try the X Challenge!" link now reads "…on Speedrun Ethereum" (one mention each, riding the existing convention).
- **Landing pages**: `/start` gains a semantic H1 (it had none; the logo is unchanged), and the `/learn-solidity` hero subcopy and `/build-prompts` intro name the brand.

No new UI elements, no keyword-changing H1 edits, no logic changes.

## Companion PRs
These complement the challenge-body changes in `scaffold-eth/se-2-challenges` (challenge READMEs render on the challenge pages):
- Prediction Markets challenge intro: https://github.com/scaffold-eth/se-2-challenges/pull/493
- DEX challenge intro: https://github.com/scaffold-eth/se-2-challenges/pull/494

PR #413, opened 2026-08-05
Self-identify challenge pages with breadcrumb structured data

denalimarsh:
## Summary

Follow-up to #401, applying the same treatment to the challenge pages.

- **Challenge pages never name Speedrun Ethereum in the page content.** The h1 reads "Challenge: Build a DEX" and stops there; the header logo is client-rendered, so crawlers that skip JavaScript see no brand at all. Added a "Speedrun Ethereum › Challenge" breadcrumb above the h1, matching the guides.
- **Challenge pages emit no `BreadcrumbList`.** They have had a `Course` node since #398, but no trail. `getChallengeStructuredData()` now returns the `Course` plus a matching `BreadcrumbList`, merged into the existing JSON-LD script.

The trail has two levels rather than the guides' three, because there is no challenges index page for a middle crumb to point at.

No heading text changes, and no changes to `CHALLENGE_METADATA`: the titles and descriptions there already name Speedrun Ethereum from #400. No layout changes beyond the one breadcrumb line.

## How to test

2. Open a challenge, e.g. `/challenge/dex`: a "Speedrun Ethereum › Challenge" breadcrumb renders above the title, and the first segment links home.
3. View source: one `ld+json` script containing both `Course` and `BreadcrumbList`, with the `BreadcrumbList` naming the challenge and its URL.
4. `yarn check-types` passes.


denalimarsh:

packages/nextjs/app/challenge/[challengeId]/page.tsx
```diff
@@ -113,6 +117,15 @@ export default async function ChallengePage(props: { params: Promise<{ challenge
               <div className="hidden lg:block" aria-hidden />
               <div className="min-w-0 flex flex-col gap-8 lg:flex-row lg:items-center lg:gap-10">
                 <div className="min-w-0 flex-1 max-w-[640px]">
+                  <nav aria-label="Breadcrumb" className="text-sm text-base-content/80 mb-3">
+                    <Link href="/" className="hover:underline">
+                      Speedrun Ethereum
+                    </Link>
+                    <span className="mx-1.5" aria-hidden>
```
Comment (carletex) [inline packages/nextjs/app/challenge/[challengeId]/page.tsx:124]: Should we replace this with the real challenge title to match the JSON-LD??

e.g. Speedrun Ethereum › Tokenization

or as in guides ("Speedrun Ethereum › Guides", but Guides links to /guides) should we link it somewhere?

Comment (denalimarsh) [reply packages/nextjs/app/challenge/[challengeId]/page.tsx:124]: Good catch, and worse than a mismatch: "Challenge" was plain text, in neither the JSON-LD nor a page that exists.

Went with the guides shape. Visible `Speedrun Ethereum › Challenges` linking to /start, and the JSON-LD adding the third item: Speedrun Ethereum → Challenges (/start) → Simple NFT Example.

My comment claiming there is no challenges index was wrong, /start calls getAllChallenges(). Fixed that too.

Your call on the label, since it reads "Challenges" but points at /start. The challenge title stays out of the visible trail because the h1 sits right under it, same as guides.

PR #418, opened 2026-08-24
Remove batch XP

rin-st:
I think we need to remove batch xp from the side quests since it's not possible to participate in batches for the new users, and hence they can't get all the xp even if they want to

image


carletex:
Thanks!

